PR #9, opened 2017-11-17
Changed PrzypisEditor flow so it accepts  `saveAction`  argument to allow form error handling within React component

kaglowka:


src/pp-annotator/form/AnnotationForm.tsx
```diff
@@ -125,15 +127,23 @@ export default class AnnotationForm extends React.Component<
...
-  private onSave() {
+  private onSave(event: any) {
     const fieldsToSave = getFormState(this.state);
     if (this.state.isLinkOnly) {
       fieldsToSave.comment = '';
     }
-    this.props.onSave(fieldsToSave);
+    this.props.annotation.fields = fieldsToSave;
+    const result = this.props.saveAction(this.props.annotation);
+    const errors = result.errors;
```
msz: This will not work when `saveAction` is an actual method that saves the annotation in the backend!

Anything interacting with the backend will not block and return results synchronously — rather, it will synchronously return a Promise with the results of the asynchronous action (communication with the backend). `saveAction` should really have a `Promise<T>` return type. This is also how the `create()` actions is defined in the Annotator [API docs](http://docs.annotatorjs.org/en/latest/api/storage.html#annotator.storage.StorageAdapter.prototype.create).

src/pp-annotator/form/PrzypisEditor.tsx
```diff
@@ -85,68 +87,57 @@ export default class PrzypisEditor extends Widget {
...
+              position: util.IPosition,
+              saveAction: (annotation: IAnnotation) => any) {
     this.annotation = annotation;
-    this.updateForm(annotation.fields || {});
+    this.saveAction = saveAction.bind(this);
+    this.updateForm(annotation);
+    this.show(position);
 
     return new Promise<IAnnotation>((resolve, reject) => {
       this.promiseResultContainer = {
```
msz: Saving Promise resolution methods like this for later resolution is a serious antipattern — it makes the flow extremely unpredictable. Since we're reworking the flow here, this might be a good place to resolve this issue.

src/pp-annotator/form/PrzypisEditor.tsx
```diff
@@ -85,68 +87,57 @@ export default class PrzypisEditor extends Widget {
...
-   * Returns an unresolved Promise that will be resolved when the save/cancel button is clicked.
-   * If load function is waited upon, it will finish only when the save/cancel button is clicked.
-   */
-  public load(annotation: annotator.IAnnotation, position: util.IPosition) {
+  public load(annotation: annotator.IAnnotation,
+              position: util.IPosition,
+              saveAction: (annotation: IAnnotation) => any) {
     this.annotation = annotation;
-    this.updateForm(annotation.fields || {});
+    this.saveAction = saveAction.bind(this);
```
msz: This doesn't seem like a good place to bind context. This class doesn't need to know what exactly `saveAction` does, so it shouldn't need to bind `this`.

kaglowka: Good point, my mistakes.

src/pp-annotator/form/AnnotationForm.tsx
kaglowka: Is what you suggest something like
```js
this.props.saveAction(this.props.annotation).then(
\\ success; close form 
).catch(
\\ render validation errors if they are the reason
)
```

or simply
```js
this.props.saveAction(this.props.annotation).then(
(result) => {
  if (result.errors)  {
\\ render validation errors
}
else {
\\ success; close form 
}

)
```

msz: This would depend on how saveAction is written, what the possible errors can be and whether they are treated as exceptional behaviour (shouldn't happen) and promise rejected, or a negative result (can happen in a normal flow) and promise resolved. So if you're going to use the method from Annotator it depends on what is does.

PR #20, opened 2017-12-28
Mock viewer

kaglowka:
- mock viewer window displayed at the app start for dev purposes
- to be
   - improved when there's a need for mocks like this
   - removed when it gets hard to maintain


config/webpack.dev.js
```diff
@@ -15,5 +15,8 @@ module.exports = merge(common, {
   plugins: [
     new webpack.NamedModulesPlugin(),
     new webpack.HotModuleReplacementPlugin(),
+    new webpack.DefinePlugin({
+      'process.env.PP_DEV_MOCK_VIEWER': true,
```
lgodziejewski: `process.env.FOO` is usually used for Node.js stuff, I think we can go with e.g. `__PP_DEV_MOCK_VIEWER__`

PR #25, opened 2018-01-08
CSS Modules

lgodziejewski:
Changes as in commit descriptions.
In short:
- fixed TSLint warnings and introduced minor config changes

- Configured css modules (a bit of PITA here - TypeScript doesn't like `import styles from 'foo.scss'`, so I had to introduce some custom loader which generates .d.ts files on the fly based on the (s)css content. I haven't commited them, as I'm not sure we want those on the repo?)

- Reworked PriorityButtons to be a Component using css modules

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,12 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@types/classnames": {
+      "version": "2.2.3",
+      "resolved": "https://registry.npmjs.org/@types/classnames/-/classnames-2.2.3.tgz",
+      "integrity": "sha512-x15/Io+JdzrkM9gnX6SWUs/EmqQzd65TD9tcZIAQ1VIdb93XErNuYmB7Yho8JUCE189ipUSESsWvGvYXRRIvYA==",
+      "dev": true
+    },
     "@types/jquery": {
       "version": "3.2.16",
       "resolved": "https://registry.npmjs.org/@types/jquery/-/jquery-3.2.16.tgz",
@@ -1914,6 +1920,7 @@
       "requires": {
         "anymatch": "1.3.2",
         "async-each": "1.0.1",
+        "fsevents": "1.1.3",
         "glob-parent": "2.0.0",
         "inherits": "2.0.3",
         "is-binary-path": "1.0.1",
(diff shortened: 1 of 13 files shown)


lgodziejewski:
I see CircleCI is failing because these `<css>.d.ts` files are missing (Webpack starts typescript compiler before these files are generated, on 2nd run it'd be OK), so maybe I should add them?

Other solution is that the custom loader will get fixed so that it is deterministic (i.e. exits after files are created, not mid-progress)

lgodziejewski:
@1 - I have fixed the class names - they should be in a quite readable form now `Component__class--[random-5-chars]`. I haven't configured that properly before as I thought that TS mangles the names, my bad 🙁 The only thing that `typings-for-css-modules-loader` does is generate typings files (`.d.ts`) for Typescript, so that it won't throw errors on `import styles from 'Bla.css'`.

@2 hard to say, basically the React parts are pretty standard, most of the complexity is on the annotator side I think. Typescript is less common, SCSS/SASS/similar are rather widely used currently. CSS Modules is IMO a very small change in how you write code, so I don't see that as a blocker.

@3 - I wouldn't call it bleeding edge. Basically [CRA](https://github.com/facebookincubator/create-react-app)  - the official "beginner's setup for React" with some tweaks (e.g. [TS version](https://github.com/Microsoft/TypeScript-React-Starter) or [with css modules and SASS](https://www.npmjs.com/package/custom-react-scripts) (and other fancy features)) give you the same. Of course, it's not common as in "the first tutorial for web stuff gave me this", but IMO for React apps it's not that exotic.

It may be hard to notice at first glance, but CSS Modules liberate you from lots of tiresome maintenance work later on (css files tend to bloat very quickly)

As for reducing the general complexity - for me the best solution would be to get rid of TS (and, if for example we'd like to still have static typing, we could introduce [Flow](https://flow.org/) - as it is far less 'demanding' than TS, works more as a linter - code is still interpreted as pure JS would be + if I remember well, syntax is 99% like TS). For example - see how we litter the code with forced typings for things such as html click events - I'm not really convinced that it improves work quality drastically. We could of course loosen up the config a bit, e.g. by removing the `no implicit any` flag :-)
\+ IMO another problem for TS is that you need typings files for every external package you'd like to use (that is, if there are no public ones, you have to write them) or else the compiler will fail (however, I guess that it's probably possible to change it in tsconfig) **HOWEVER** - if I see correctly, turning off the "strict mode" gets rid of this problem.

I mean, I don't want to bash Typescript in general, maybe just the config is a bit too strict 😉

kaglowka:
Thank you for your exhaustive answer.

@1: The new class name generation looks great.
@2 & @3:
Well, I'm sure you know better about CSS and besides I agree CSS modules can nicely organize our styles.

As for Typescript expenses, It's true, I spend a bit more time than I wish adding typings that I find not important enough (elements that play no bigger part or "legacy" annotator parts that will be removed soon and are temporarily typed only for TS to be happy).
I think we could be happy with less strict, more optional Typescript. Maybe let's keep [Flow](https://flow.org/) in mind, but for the time being, try to configure TS to our liking?

I approved the PR but I think we should draw an agenda for the future work (in the meantime) to improve some of the current configuration.
- improve CSS modules to pass CI checks
- configure TS to suit our needs

PR #26, opened 2018-01-11
Storage changes

lgodziejewski:
Added logger and other minor changes to `DebugStorage`


kaglowka:
Thanks for fixing! (I was wondering what caused the sudden appearance of zeros...)

PR #31, opened 2018-01-15
Loading annotations from a .json file (set in app-settings)

kaglowka:
- configured json loading for TS
- temporary dev optional functionality of loading annotations from a `.json` file (set in `app-settings`) for UX researchers to be able to freely test the app before we finish the data layer setup


src/index.ts
```diff
@@ -12,37 +12,72 @@ import 'css/editor.scss';
...
+
+function attachAnnotationsFromFile(annotatorApp: App) {
+  // Reassign annotations' ids starting from 1000
+  // so they don't collide with the ids of the annotations created through the application
+  let id = 1000;
+  for (const annotation of input_annotations) {
+    if (annotation.url === window.location.href) {
+      annotation.id = id;
+      annotatorApp.annotations.create(annotation);
+      id += 1;
```
lgodziejewski: maybe instead of overwriting data here, provide some very big random ids in the `json` file?

src/pp-annotator/api/DebugStorage.ts
```diff
@@ -14,6 +17,11 @@ export default class DebugStorage implements IStorage, IModule {
 
   trace(action: string, data: any) {
     console.debug(`DebugStorage: invoking action ${action} on data: `, data);
+    // If some annotations are hardcoded, print an annotation json to facilitate copying it manually to the file
+    if (PP_SETTINGS.READ_ANNOTATIONS_FROM_FILE) {
+      console.debug('Annotation json:');
+      console.debug(JSON.stringify(data));
```
lgodziejewski: Now every time something is done (and the env var is set), this JSON is printed. Is this intentional?

tsconfig.json
```diff
@@ -24,6 +24,7 @@
   },
   "include": [
     "./@types/styles.d.ts",
+    "./@types/json.d.ts",
```
lgodziejewski: maybe it'd be easier to just have a `custom.d.ts` file with everything inside of it (css, scss and json and any other future stuff)?

@types/json.d.ts
```diff
@@ -0,0 +1,5 @@
+// Typing necessary to load .json files in Typescript
+declare module "*.json" {
+    const value: any;
+    export default value;
+}
```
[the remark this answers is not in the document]
lgodziejewski: (+ you could do it the shorthand way, as in the `styles` file )

src/index.ts
```diff
@@ -12,37 +12,72 @@ import 'css/editor.scss';
 import IPPSettings from './PPSettings.interface';
 declare const PP_SETTINGS: IPPSettings;
 
+// Optional external hardcoded annotations
+import input_annotations from '../config/annotations.json';
```
lgodziejewski: do we have a `json-loader` in webpack config or is it handled by one of the currently used ones?

kaglowka: Webpack 2+ has a built-in json loader ;)

src/pp-annotator/api/DebugStorage.ts
kaglowka: Yes (I should've made a comment in the PR description)

As a substitute to a database, to create a persistent annotation in the application (for demo purposes), you can copy the json from the console output, paste it into `annotations.json` and there you are ;)

src/index.ts
kaglowka: The point is to let anybody add freely their own annotation, without having to worry about ids (and manually generate a random id).

This is a short-lived solution and behaves nicely

tsconfig.json
kaglowka: Good point

PR #39, opened 2018-01-30
Viewer and editor UX/UI improvements + browser extension CSS "scope limitation".

kaglowka:
### styles
- major CSS refactor (clearer functional division between files)
- more selective setup od Semantic UI CSS (no global styles, only selected components imported)
- our custom `pp-ui` CSS base class that limits our styles' influence on other websites

### UI
- priority button & link field tooltips
- better pointer/highlight UX
    - fixed at least one source of viewer-not-disappearing bug
    - decreased viewer disappearance latency
    - stopped viewer from wobbling when the user shakes the pointer over the highlighted text
- relative date display
- `does_belong_to_user` field to control user permissions
- annotation delete confirmation window
- some small text, spacing and style changes to editor and viewer windows


assets/three-dots.svg
```diff
@@ -0,0 +1 @@
+<?xml version="1.0" ?><svg id="Layer_1_1_" style="enable-background:new 0 0 16 16;" version="1.1" viewBox="0 0 16 16" xml:space="preserve" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><circle cx="2" cy="8" r="2"/><circle cx="8" cy="8" r="2"/><circle cx="14" cy="8" r="2"/></svg>
```
lgodziejewski: if you want to cut size a bit you can use [SVGOMG](https://jakearchibald.github.io/svgomg/)

src/css/common/shame.scss
```diff
@@ -0,0 +1,10 @@
+/*
+  Shameful overrides
+ */
+
+//semantic ui dimmer
+// WARNING: this is a temporary obviously fallacious setting that will influence dimmers on
+// ALL websites using semantic ui...
+.ui.dimmer {
```
lgodziejewski: what's the reason it can't be scoped too?

src/pp-annotator/api/Annotation.interface.ts
```diff
@@ -1,4 +1,5 @@
 export default interface IAnnotation {
   id: number;
-  data: any;
+  // KG removed it, as I see no purpose of this field
+  // data: any;
```
lgodziejewski: I'd propose removing this comments

src/pp-annotator/api/DebugStorage.ts
```diff
@@ -24,36 +24,38 @@ export default class DebugStorage implements IStorage, IModule {
     }
   }
 
-  create(annotation: IAnnotation): IAnnotation {
+  create(annotation: IAnnotationAPIModel): IAnnotationAPIModel {
     if (!annotation.id) {
       annotation.id = generateId();
     }
+    annotation.create_date = new Date(Date.now()).toISOString();
```
lgodziejewski: I think `new Date()` by default has the value of `Date.now()`. The latter just returns a `number` instead of `Date object`

src/pp-annotator/api/DebugStorage.ts
```diff
@@ -24,36 +24,38 @@ export default class DebugStorage implements IStorage, IModule {
...
-
-    return deleted;
+    return deleted[0];
   }
 
-  query(id: string | number): IAnnotation {
-    this.trace('query', id);
-
-    return this.annotations.find(annotation => annotation.id === id) as IAnnotation;
+  query(id?: number): IAnnotationAPIModel[] {
```
lgodziejewski: why the change?

src/pp-annotator/api/Storage.interface.ts
```diff
@@ -4,5 +4,5 @@ export default interface IStorage {
   create(annotation: IAnnotation): IAnnotation;
   update(annotation: IAnnotation): IAnnotation;
   delete(annotation: IAnnotation): IAnnotation;
-  query(id: number): IAnnotation;
+  query(query?: any): IAnnotation[];
```
lgodziejewski: If I see correctly the interface is out of sync with the changes above

src/pp-annotator/editor/EditorContent.scss
```diff
@@ -2,9 +2,10 @@
 
   .header-buttons {
     display: flex;
+    justify-content: space-between;
 
-    :not(:first-child):not(:last-child) {
-      margin: 0 12px;
+    button {
+      width: 32%;
```
lgodziejewski: `flex-grow: 1` should be enough, I think?
setting fixed width inside `flex` containers is usually a "last resort" solution 😛

src/pp-annotator/editor/EditorContent.tsx
```diff
@@ -158,35 +159,30 @@ export default class EditorContent extends Component<IEditorContentProps, IEdito
...
+            >
+              {annotationPrioritiesLabels.NORMAL}
             </PriorityButton>
-            <PriorityButton type={AnnotationPriorities.WARNING} onClick={this.setPriority} priority={priority}>
-              wyjaśnienie
+            <PriorityButton
+              type={AnnotationPriorities.WARNING}
+              onClick={this.setPriority}
+              priority={priority}
+              tooltipText="Bez tego przypisu czytelnik może być wprowadzony w&nbsp;błąd"
```
lgodziejewski: why the `&nbsp;` ?

src/pp-annotator/modules/PPUI.ts
```diff
@@ -122,100 +122,102 @@ interface IState {
...
   start(app: App) {
+    let state: IState;
     const parseRanges = this.parseRanges;
+    /*
+      State object declarations
+     */
+    const highlighter = new annotator.ui.highlighter.Highlighter(this.element);
+    const textselector = new annotator.ui.textselector.TextSelector(this.element, {
+      onSelection(ranges, event) {
+        if (!state) {
```
lgodziejewski: if you'd write this function as `onSelection: (ranges, event) => { ... },` it should be possible to just use `this.state`, instead of passing a dedicated `state` here.

Unless `this` is required here to be the default bind (however I see no usage of it?)

src/pp-annotator/viewer/ViewerContentItem.tsx
```diff
@@ -1,5 +1,9 @@
 import React from 'react';
 import AnnotationViewModel from '../annotation/AnnotationViewModel';
+import {AnnotationPriorities, annotationPrioritiesLabels} from '../consts';
+import {Popup, Modal, Button} from 'semantic-ui-react';
+import moment from 'moment';
+import 'moment/locale/pl';
```
lgodziejewski: I guess it's better to put this in the `vendor.ts` file

src/css/common/shame.scss
kaglowka: There is no obvious way to set our own a class for the semantic ui dimmer appearing together with modal (it might be possible to set it if you separately display a dimmer and a modal, but then again I guess it's not the way you're supposed to do it)

src/pp-annotator/api/Annotation.interface.ts
kaglowka: agreed

src/pp-annotator/api/DebugStorage.ts
kaglowka: That's right (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date), I'll change that.

src/pp-annotator/api/DebugStorage.ts
kaglowka: If you mean the `id`'s optionality in `query` method, the reason is to provide an intuitive way to get all/selected annotations.
If you mean the overall change: I needed to give up on the `DebugStorage's` universality and simulate logic on an annotation's creation (author saving + creation date). In the same vein, I gave up on `string | number` alternative (I knew we would not provide a string any more and didn't want to maintain that).

src/pp-annotator/api/Storage.interface.ts
kaglowka: Corrected

src/pp-annotator/editor/EditorContent.tsx
kaglowka: To prevent a line break (just found out you can use &nbsp; for this purpose)

src/pp-annotator/viewer/ViewerContentItem.tsx
kaglowka: agreed

src/pp-annotator/modules/PPUI.ts
kaglowka: I actually did it this way, but after the modification I thought the readability was much worse afterwards and decided to keep the intermediary state variable. Not sure if you see it this way, too.
It's not a big deal, but I find lines like 
```js
this.state.menu.load(annotation, this.state.interactionPoint);
```
a bit ugly. The reason is there are no other references to `this` in all callbacks like `onSelection` except in `this.state`, and, probably, by design there should be no other references to the `PPUI` instance; only references to other components (which is a flow we should probably change, but then better make a radical change then have a mixture of different approaches) . If so, `this` is  a redundant prefix to `state`.

This is probably more words than it's really worth ;) and well, this is just my feeling about current application flow and this issue remains more of a philosophical issue, as we will change the flow soon anyway.

A full version after the modification, so you can judge yourself:
```js
 const highlighter = new annotator.ui.highlighter.Highlighter(this.element);
    const textselector = new annotator.ui.textselector.TextSelector(this.element, {
      onSelection(ranges, event) {
        if (!this.state) {
          throw new Error('App not initialized!');
        }

        if (!this.state.editor.isShown() && ranges.length > 0) {
          const url = window.location.href;
          const annotation = AnnotationViewModel.fromSelection(parseRanges(ranges), url);
          this.state.interactionPoint = util.mousePosition(event);
          this.state.menu.load(annotation, this.state.interactionPoint);
        } else {
          this.state.menu.hide();
        }
      },
    });
```

lgodziejewski: maybe inside of `onSelection` deconstruct `this.state`, e.g.:
```
const {
  editor,
  menu,
  interactionPoint,
} = this.state;
```
that would remove the issue you've described (bloated `this.state.foo.bar`) still keeping the changes I've suggested 🙂

PR #46, opened 2018-04-16
Show menu on select, rework editor reducer

lgodziejewski:
Introduced changes as in commits:
- added [Redux DevTools](https://github.com/zalmoxisus/redux-devtools-extension) config, so that redux store can be easier debugged
- reworked `editor` reducer to handle all widget states
- renamed `selector` reducer to `textSelector` to be more specific
- added dispatch `showMenu` in onSelection callback with calculated mouse position (needs to be tweaked a bit as the position isn't very accurate)


src/store/widgets/selectors.ts
```diff
@@ -0,0 +1,22 @@
+import { createSelector } from 'reselect';
+import { IStore } from 'store/reducer';
+
+function selectWidgetState({ inverted, location, visible }) {
```
kaglowka: Does this selector play any other role than a pure function would? (as far as I understand the purpose of selectors is to cache derived state) Does it make sense to introduce selectors before you suspect any performance issues?

lgodziejewski: no calculations here, so no performance gain - that is true.

PR #50, opened 2018-04-21
UI restore

### LINKED ISSUE #33: Production build error
`npm run build` produces
```
ERROR in main.bundle.js from UglifyJs
Name expected [main.bundle.js:1,338684]
```

kaglowka:
**Fixes**:
- flattened `Editor` props and created editor selectors so it reacts only to actual prop changes
- finally limited `DraggableWidget` responsibilities to basic callback setup (before the modification every `Editor` input change reset its location to the one before dragging.
- fixed `Highlighter` to work with overlapping annotations
- restored prod build for GitHub CI

**New** (in Redux architecture)
- `Menu` disappears on deselection
- `Editor` pops up on `Menu` click
- `Viewer` appears on highlight mouseover
- temporary `state.annotations` further development to mock the backend before we connect to it
- `Highlights` component connected to the store that creates text highlights based on `state.annotations` (using `Highlighter` util)
- `Viewer`'s `connect` and selectors

**To Do**
- [ ] refactor after the changes
- [ ] `Viewer` disappearing on mouseleave

Fixes #33

### DIFF
--- a/src/components/editor/Editor.tsx
+++ b/src/components/editor/Editor.tsx
@@ -13,6 +13,7 @@ import {IEditorState, IEditorProps, IEditorForm} from './interfaces';
 import { DraggableWidget } from 'components/widget';
 import { hideEditor } from 'store/widgets/actions';
 import {createAnnotation} from '../../store/annotations/actions';
+import {Range} from 'xpath-range';
 
 @connect(
   (state) => {
@@ -20,6 +21,7 @@ import {createAnnotation} from '../../store/annotations/actions';
       visible,
       locationX,
       locationY,
+      range,
       // form
       annotationId,
       priority,
@@ -32,6 +34,7 @@ import {createAnnotation} from '../../store/annotations/actions';
       visible,
       locationX,
       locationY,
+      range,
       // form
       annotationId,
       priority,
@@ -42,8 +45,11 @@ import {createAnnotation} from '../../store/annotations/actions';
   },
   dispatch => ({
     hideEditor: () => dispatch(hideEditor()),
-    createAnnotation: (form: IEditorForm) => {
-      dispatch(createAnnotation(form));
+    createAnnotation: (form: IEditorForm, range: Range.SerializedRange) => {
+      dispatch(createAnnotation({
+        ...form,
+        range,
+      }));
     },
   }),
 )
@@ -170,7 +176,7 @@ class Editor extends React.Component<
   save() {
     // TODO provide real query promise; for now just a placeholder (createAnnotation is to be removed)
     const fetchData = new Promise((resolve, reject) => {
-      this.props.createAnnotation(this.state);
+      this.props.createAnnotation(this.state, this.props.range);
       resolve();
     });
 
--- a/src/components/editor/Editor.tsx
+++ b/src/components/editor/Editor.tsx
@@ -66,6 +66,7 @@ class Editor extends React.Component<
       referenceLink: nextProps.referenceLink,
       referenceLinkTitle: nextProps.referenceLinkTitle,
 
+      moved: false,
       locationX: nextProps.locationX,
       locationY: nextProps.locationY,
       referenceLinkError: '',
@@ -84,6 +85,15 @@ class Editor extends React.Component<
 
   isNewAnnotation() {
     return this.props.annotationId !== null;
+  }
+
+  onDrag = (delta: IVec2) => {
+    this.setState({
+      locationX: this.state.locationX + delta.x,
+      locationY: this.state.locationY + delta.y,
+      moved: true,
+    });
+    return true;
   }
 
   setPriority = (priority: AnnotationPriorities) => {
@@ -181,6 +191,7 @@ class Editor extends React.Component<
     const {
       locationX,
       locationY,
+      moved,
       priority,
       comment,
       referenceLink,
@@ -200,6 +211,8 @@ class Editor extends React.Component<
         locationX={locationX}
         locationY={locationY}
         mover={this.moverElement}
+        onDrag={this.onDrag}
+        calculateInverted={!moved}
       >
         <div className={styles.headBar}>
           <label className={styles.priorityHeader}> Co dodajesz? </label>
--- a/src/components/editor/Editor.tsx
+++ b/src/components/editor/Editor.tsx
@@ -176,7 +176,7 @@ class Editor extends React.Component<
   save() {
     // TODO provide real query promise; for now just a placeholder (createAnnotation is to be removed)
     const fetchData = new Promise((resolve, reject) => {
-      this.props.createAnnotation(this.state, this.props.range);
+      this.props.createAnnotation(this.state as IEditorForm, this.props.range);
       resolve();
     });
 
--- a/src/components/editor/Editor.tsx
+++ b/src/components/editor/Editor.tsx
@@ -1,19 +1,17 @@
 import React, {RefObject} from 'react';
 import {connect} from 'react-redux';
 import classNames from 'classnames';
-import {selectEditorState} from 'store/selectors';
-
-import {AnnotationPriorities, annotationPrioritiesLabels} from '../consts';
-import styles from './Editor.scss';
-import {DragTracker, IVec2} from '../../utils/move';
-import PriorityButton from './priority-button/PriorityButton';
+import {Range} from 'xpath-range';
 import {Modal, Popup} from 'semantic-ui-react';
-import Widget from '../widget/Widget';
+import {AnnotationPriorities, annotationPrioritiesLabels} from '../consts';
+import {DragTracker, IVec2} from 'utils/move';
+import PriorityButton from './priority-button/PriorityButton';
 import {IEditorState, IEditorProps, IEditorForm} from './interfaces';
 import { DraggableWidget } from 'components/widget';
-import { hideEditor } from 'store/widgets/actions';
-import {createAnnotation} from '../../store/annotations/actions';
-import {Range} from 'xpath-range';
+import { hideEditor, createAnnotation} from 'store/actions';
+import {selectEditorState} from 'store/selectors';
+
+import styles from './Editor.scss';
 
 @connect(
   (state) => {
--- a/src/components/editor/Editor.tsx
+++ b/src/components/editor/Editor.tsx
@@ -5,74 +5,46 @@ import {selectEditorState} from 'store/selectors';
 
 import {AnnotationPriorities, annotationPrioritiesLabels} from '../consts';
 import styles from './Editor.scss';
+import {DragTracker, IVec2} from '../../utils/move';
 import PriorityButton from './priority-button/PriorityButton';
 import {Modal, Popup} from 'semantic-ui-react';
-import {DraggableWidget} from 'components/widget';
-
-interface IEditorProps {
-  visible: boolean;
-  locationX: number;
-  locationY: number;
-  form: IEditorForm;
-  editor: any;
-}
-
-export interface IEditorForm {
-  priority: AnnotationPriorities;
-  comment: string;
-  referenceLink: string;
-  referenceLinkTitle: string;
-}
-
-export interface IEditorState extends IEditorForm {
-  locationX: number;
-  locationY: number;
-  isDragged: boolean;
-  referenceLinkError: string;
-  referenceLinkTitleError: string;
-  noCommentModalOpen: boolean;
-}
-
-function annotationForm(annotation?): IEditorForm {
-  const model: any = annotation || {};
-  return {
-    priority: model.priority || AnnotationPriorities.NORMAL,
-    comment: model.comment || '',
-    referenceLink: model.referenceLink || '',
-    referenceLinkTitle: model.referenceLinkTitle || '',
-  };
-}
+import Widget from '../widget/Widget';
+import {IEditorState, IEditorProps} from './interfaces';
+import { DraggableWidget } from 'components/widget';
 
 @connect((state) => {
-  let form;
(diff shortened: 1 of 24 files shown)


src/components/menu/Menu.tsx
```diff
@@ -25,8 +32,14 @@ interface IMenuProps {
     visible,
     locationX,
     locationY,
+    range: state.textSelector.range,
   };
-})
+},
+  dispatch => ({
```
lgodziejewski: if you just want to wrap actions into dispatch, it's possible to just pass an object here, like:
```
{
  showEditor: showEditorNewAnnotation,
  hideMenu,
}
```

`connect` automatically wraps them in `dispatch` (not sure about the 1st one, but I guess it passes arguments to the wrapped function)

src/components/editor/Editor.tsx
```diff
@@ -5,74 +5,62 @@ import {selectEditorState} from 'store/selectors';
 
 import {AnnotationPriorities, annotationPrioritiesLabels} from '../consts';
 import styles from './Editor.scss';
+import {DragTracker, IVec2} from '../../utils/move';
 import PriorityButton from './priority-button/PriorityButton';
 import {Modal, Popup} from 'semantic-ui-react';
-import {DraggableWidget} from 'components/widget';
+import Widget from '../widget/Widget';
+import {IEditorState, IEditorProps, IEditorForm} from './interfaces';
+import { DraggableWidget } from 'components/widget';
```
lgodziejewski: `import Widget` and this one can be merged into one import - `import Widget, { DraggableWidget }` ... `

src/components/editor/Editor.tsx
```diff
@@ -5,74 +5,62 @@ import {selectEditorState} from 'store/selectors';
...
+import { hideEditor } from 'store/widgets/actions';
```
lgodziejewski: this should be importable from `'store/actions'`, along with the one below

src/components/editor/Editor.tsx
```diff
@@ -106,7 +100,16 @@ class Editor extends React.PureComponent<Partial<IEditorProps>,
   }
 
   isNewAnnotation() {
-    return this.props.editor.annotationId !== null;
+    return this.props.annotationId !== null;
```
lgodziejewski: 👍

src/components/editor/Editor.tsx
```diff
@@ -106,7 +100,16 @@ class Editor extends React.PureComponent<Partial<IEditorProps>,
...
+  }
+
+  onDrag = (delta: IVec2) => {
```
lgodziejewski: Couldn't this be handled inside `DraggableWidget` ?

src/components/menu/Menu.tsx
```diff
@@ -39,9 +52,15 @@ export default class Menu extends React.Component<Partial<IMenuProps>, {}> {
...
-    console.log(event);
+  onClick = (e: any) => {
+    const {
+      locationX,
+      locationY,
+      range,
+    } = this.props;
+
+    this.props.hideMenu();
+    this.props.showEditor(locationX, locationY, range);
```
lgodziejewski: 👍

src/components/viewer/Viewer.tsx
```diff
@@ -44,10 +58,9 @@ export default class Viewer extends React.Component<Partial<IViewerProps>, {}> {
       <Widget
         className={classNames('pp-ui', styles.self)}
         visible={this.props.visible}
-        invertedX={this.props.invertedX}
-        invertedY={this.props.invertedY}
         locationX={this.props.locationX}
         locationY={this.props.locationY}
+        calculateInverted={true}
```
lgodziejewski: It should be possible to just write `calculateInverted` here, unless TS has a problem with this 😛

src/components/widget/DraggableWidget.tsx
```diff
@@ -68,21 +62,12 @@ export default class DraggableWidget extends React.PureComponent<
...
-    } = this.state;
-
     return (
       <Widget
         {...this.props}
-        locationX={locationX}
-        locationY={locationY}
-        calculateInverted={!moved}
       >
         {this.props.children}
```
lgodziejewski: side note: this is already passed in `{...this.props}`, so `<Widget {...this.props} />` should work fine

src/core/bootstrap.ts
```diff
@@ -2,9 +2,11 @@ import { Range } from 'xpath-range';
 
 import { mousePosition } from 'common/dom';
 import store from 'store';
-import { showMenu, textSelectedAction } from 'store/actions';
+import { showMenu, makeSelection } from 'store/actions';
 
 import { Highlighter, TextSelector } from './index';
+import {hideMenu} from 'store/widgets/actions';
```
lgodziejewski: I'd propose to move these two imports to the one in line 5

src/store/annotations/reducers.ts
```diff
@@ -1,6 +1,23 @@
...
+
+function maxId(state) {
+  return Math.max(0, ...state.data.map(annotation => annotation.annotationId)) + 1;
+}
 
 export default function editor(state = initialState, action) {
-  return state;
+  switch (action.type) {
+    case CREATE_ANNOTATION:
+      return Object.assign({}, {
```
lgodziejewski: I'd propose the following, a bit simpler and future-proof (won't need to modify it if any new fields gets added to this reducer)
```
return {
  ...state,
  data: state.data.concat({
    ...action.payload,
    annotationId: maxId(state),
  });
};
```

src/store/annotations/reducers.ts
```diff
@@ -1,6 +1,23 @@
...
-const initialState = [];
+const initialState = {
+  data: [],
+};
+
+function maxId(state) {
+  return Math.max(0, ...state.data.map(annotation => annotation.annotationId)) + 1;
+}
 
 export default function editor(state = initialState, action) {
```
lgodziejewski: As you've named the array field `data`, I'd suggest renaming this function to `annotations`

src/store/annotations/reducers.ts
```diff
@@ -1,6 +1,23 @@
+import {CREATE_ANNOTATION} from './actions';
 
-const initialState = [];
+const initialState = {
+  data: [],
```
lgodziejewski: 👍

src/store/annotations/reducers.ts
lgodziejewski: and maybe move it to `createAnnotationActionHandler`, similarly as in `textSelector/reducers.ts` ?

src/store/widgets/reducers.ts
```diff
@@ -62,3 +75,14 @@ function menuActionHandler(state, payload) {
     },
   };
 }
+
+// todo here: update only when the annotations have changed
```
lgodziejewski: this would probably require splitting `annotationIds` update to a separate action type.

On the other hand, if the `annotationIds` passed here is the same(by ref) array as in previous state, it should be fine

src/test.html
```diff
@@ -3,7 +3,7 @@
 <head>
     <meta charset="utf-8">
     <title><%= htmlWebpackPlugin.options.title %></title>
-    <link href="//fonts.googleapis.com/css?family=Open+Sans:300,600" rel="stylesheet">
+    <link href="http://fonts.googleapis.com/css?family=Open+Sans:300,600" rel="stylesheet">
```
lgodziejewski: lack of protocol makes the browser use the current protocol from the address bar.

Was there a problem with this?

src/components/highlights/Highlights.ts
```diff
@@ -0,0 +1,59 @@
...
+
+@connect(
+  state => ({
+    annotations: state.annotations.data,
+  }),
+  dispatch => ({
+    showViewer: (x, y, annotationId) => dispatch(showViewer(x, y, annotationId)),
+  }),
+)
+export default class Highlights extends React.Component<Partial<IHighlightsProps>, {}> {
```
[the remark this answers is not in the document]
kaglowka: I wrapped this logic in a React component just to comply with Redux philosophy (and more prosaically, to use `connect`). I guess we can also use `connect` more directly or connect to the store in any other way. The important point was to render annotations and rerender them if they change.

src/components/menu/Menu.tsx
kaglowka: Good to know, I'll check

src/components/editor/Editor.tsx
kaglowka: Thanks, I missed this particular "shortcut" pattern you made.

src/components/viewer/Viewer.tsx
kaglowka: Actually it does :D or at least it had in some of our early default configurations

src/store/widgets/reducers.ts
kaglowka: Oh, that was an unclear comment.
What I meant was that location should not be changed unless the viewer's annotations have changed -- a simple solution I used in the previous architecture that stopped the window from changing location every time the cursor slipped off the window for a split second

src/test.html
kaglowka: It only didn't in the `file:\\\' protocol. I fixed it not wondering if we'd support the `file:///` environment.

src/components/editor/Editor.tsx
lgodziejewski: Wouldn't it be enough to:
- keep `locationX` and `locationY` as `DraggableWidget` state (btw. I think this could be squashed to `location: { x, y }`, as any change of either x or y must result in a rerender - and we won't have to pass two fields through some of the components who don't care about them)
- move the `deriveStateFromProps` logic to the constructor (so it's only set when `DraggableWidget` is created - from the initial `x` and `y` coords passed from `Editor` or any other future component)
- (optional, I guess?) remove `componentDidUpdate()` from `DraggableWidget` (unless recreating `dragTracker` is necessary, but I highly doubt it?)

src/components/highlights/Highlights.ts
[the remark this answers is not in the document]
lgodziejewski: I think [store.subscribe](http://devdocs.io/redux/api/store#subscribe) would me perfectly suitable for this scenario.

it could be handled in a plain ts module(file) imported and initialized in `bootstrap.ts`, for example (or if short enough, directly in `bootstrap.ts`)

src/test.html
lgodziejewski: Ha, didn't think about that 😛

src/store/widgets/reducers.ts
lgodziejewski: You mean a scenario when a user hovers over the highlight, which results in the `Viewer` window getting moved by a couple of pixels every time?

So to avoid it - if annotation id is same as before, only toggle visibility (without modifying the { x, y } location) ?

src/components/editor/Editor.tsx
lgodziejewski: The `store/actions.ts` file just re-exports stuff from internal actions.ts files, similarly `store/selectors.ts` 🙂

kaglowka: Yes, that's what I meant by "shortcut" :) I just didn't notice this pattern before.

src/components/editor/Editor.tsx
kaglowka: Widget location is kept in the global `state`. If it changes, it's passed to Editor. How can we keep Editor from overwriting (by passing on the location) the location DraggableWidget tries to maintain?

src/store/widgets/reducers.ts
kaglowka: Precisely :) (i'll update the comment)

src/components/menu/Menu.tsx
kaglowka: I tried with `hideMenu` and it stops working. Are you sure about this shortcut?

lgodziejewski: like so?
```
@connect( state => ({
  ...
}), {
  hideMenu,
  otherFn,
})
```

src/components/editor/Editor.tsx
lgodziejewski: as discussed on Slack - if it's the first render of `DraggableWidget` - only `constructor` (and later `componentDidUpdate`, but constructor will be fine I think) will be invoked.
If it's a further update - `componentDidUpdate` will be invoked.

However for this to work, I'd propose to move the `visible` flag handling from `Widget` to `Editor` (a simple `if (!visible) return null` would be sufficient at the beginning of render fn) - so that `DraggableWidget` is actually recreated each time visible is changed from false to true.

On second thought, actually dismounting the whole `Editor` component and mounting it later on according to `visible` status would be even better, but then we'd have to introduce some kind of `WidgetManager`, which would handle stuff like this (i.e. conditional rendering of `Editor`, `Viewer`, `Menu` etc, instead of them being always render in `App`)

src/components/highlights/Highlights.ts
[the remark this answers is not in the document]
kaglowka: Moved to roadmap (`miscalleneous`)

src/components/menu/Menu.tsx
kaglowka: Like so:
```
showEditor: (x, y, range) => dispatch(showEditorNewAnnotation(x, y, range)),
hideMenu,
```
and
```
showEditor: (x, y, range) => dispatch(showEditorNewAnnotation(x, y, range)),
hideMenu: hideMenu(),
```

src/components/editor/Editor.tsx
kaglowka: Moved to roadmap (`miscallaneous`)

PR #52, opened 2018-05-03
WPI: Api connect

glowka:


package.json
```diff
@@ -14,6 +14,7 @@
     "react-dom": "^16.3.1",
     "react-redux": "^5.0.7",
     "redux": "^3.7.2",
+    "redux-json-api": "^2.0.0-beta.5",
```
kaglowka: It's nice to already use 2.0 beta version :)

glowka: Yeah, actually it is pretty stable ;)

PR #53, opened 2018-05-08
WIP: Further api connect

kaglowka:
- restructured application initialization
- typed annotation API usage with AnnotationAPIModel & AnnotationAPIPostModel
- updated Editor/Viewer field names to match API
- connected Viewer to API model


src/api/annotations.ts
```diff
@@ -0,0 +1,36 @@
+import {AnnotationPriorities} from 'components/consts';
+
+export interface RangeAPIModel {
+  start: string;
+  startOffset: number;
+  end: string;
+  endOffset: number;
+}
+
+export interface AnnotationAPIPostModel {
```
glowka: I would say we at least one abstraction layer away from that HTTP, so what about referring to model/resource life cycle and using _create_ or something like that?

src/components/highlights/Highlights.ts
```diff
@@ -32,16 +34,16 @@ export default class Highlights extends React.Component<Partial<IHighlightsProps
...
 
   drawAll() {
-    // For each annotation, it is cleared and drawn again
+    // For each annotation, it is cleared and redrawn
     this.highlighter.drawAll(this.props.annotations.map(annotation => ({
-      id: annotation.annotationId || annotation.id,
+      id: annotation.id,
       range: annotation.attributes.range,
-      annotationData: annotation.attributes,
+      annotationData: annotation,
```
glowka: Yeah, much cleaner now

src/index.tsx
```diff
@@ -15,35 +16,17 @@ import './css/common/pp-semantic-ui-reset.scss';
...
-  // TODO: 2. this url is hardcoded now, but should be imported from config
-  store.dispatch(setAxiosConfig({
-    baseURL: 'http://localhost:8000/api',
-  }));
-
-  // This our root request that need to have part of the url (path) hardcoded
+function loadData() {
+  // This is our root request that needs to have part of the url (path) hardcoded
   store.dispatch(readEndpoint('/annotations?url=' + getAnnotationUrl()));
-
```
glowka: `getAnnotationUrl` can be safely replaced with normal,raw and universal `window.location` thanks to changes from incoming https://github.com/PrzypisPowszechny/pp/pull/16 which will handle full urls. Both when creating annotation and then when listing annotation with filtering query (`?url=` ). The original, full url will be stored as well as processed version used for efficient and accurate queries, with the results based on the relevant part of url only (without protocol etc).

src/index.tsx
```diff
@@ -15,35 +16,17 @@ import './css/common/pp-semantic-ui-reset.scss';
...
-    documentContainer,
-  );
 }
 
 const isBrowser = typeof window !== 'undefined';
 if (isBrowser) {
   initializeCoreHandlers();
-  injectApp();
+  injectComponents();
+  loadData();
```
glowka: How about referring to _init_? It would be nice to have it in that name as it is that one particular data loading. My first try is `loadInitialData`, but that's just a try...

src/store/store.ts
```diff
@@ -16,4 +17,8 @@ const store = createStore(
   applyMiddleware(thunk, promise, logger),
 );
 
+store.dispatch(setAxiosConfig({
+  baseURL: 'http://localhost:8000/api',
```
glowka: Do you have any idea where to put the proper config to make it flexible for the development and production?

src/api/annotations.ts
kaglowka: Actually I didn't mean HTTP POST method (but the *post* action) but I agree that `create` is better, I'll change it

src/index.tsx
kaglowka: I'll do it right away

src/index.tsx
kaglowka: :+1:

src/store/store.ts
kaglowka: I see two options:
1) For now we have two webpack config files for dev and prod. Maybe just put there the URL as a variable/config key until more webpack-independent settings appear on the horizon when it makes sense to create more config files?
2) use `config/app-settings.js`.

I first thought of option 1 but now I'm more into option 2. How about it?

src/api/annotations.ts
glowka: OK

Off-topic: what's the difference between post action and method for you here? What do you mean by _action_ ?

src/store/store.ts
glowka: After considering few options I would go for 2 too. How about setting it based on `NODE_ENV` (`process.env.NODE_ENV` to be exact)?

This would require adding something like this in webpack plugins to expose it in built code:
```
new webpack.DefinePlugin({
      'process.env': ...  // 'development' or 'production' or whatever environment
    }),
```

config/app-settings.js
```diff
@@ -1,8 +1,8 @@
 exports.appSettings = {
   development: {
-    // for now empty, but used before and used in the future :)
+    API_URL: "http://localhost:8000/api",
   },
   production: {
-    // for now empty, but used before and used in the future :)
+    API_URL: "",
```
glowka: Why not add http://przypispowszechny.pl/api now?

webpack.config.js
```diff
@@ -149,8 +149,8 @@ const config = (env, argv) => ({
       filename: 'index.html',
     }),
     new webpack.DefinePlugin({
-      // use appropriate (dev or production) PP settings
-      PP_SETTINGS: appSettings[argv.mode]
+      // use appropriate (development or production) PP settings
+      PP_SETTINGS: JSON.stringify(appSettings[argv.mode]),
```
glowka: Oh, I haven't noticed that and was thinking about something a bit different, but it doesn't matter. It does the job well!

src/store/store.ts
```diff
@@ -16,4 +17,9 @@ const store = createStore(
   applyMiddleware(thunk, promise, logger),
 );
 
+console.log(PP_SETTINGS);
```
glowka: ups

PR #54, opened 2018-05-08
Editor further data flow

kaglowka:
- `Viewer` opening on edit button click
- annotation update with `Editor`

### DIFF
--- a/src/components/editor/Editor.tsx
+++ b/src/components/editor/Editor.tsx
@@ -1,6 +1,6 @@
 import React, {RefObject} from 'react';
 import {connect} from 'react-redux';
-import { createResource } from 'redux-json-api'
+import { createResource, updateResource } from 'redux-json-api';
 import classNames from 'classnames';
 import {Range} from 'xpath-range';
 import {Modal, Popup} from 'semantic-ui-react';
@@ -45,7 +45,13 @@ import {AnnotationAPICreateModel} from 'api/annotations';
   },
   dispatch => ({
     hideEditor: () => dispatch(hideEditor()),
-    createAnnotation: (data: AnnotationAPICreateModel) => dispatch(createResource(data)),
+    createAnnotation: (model: AnnotationAPICreateModel) => {
+      if (model.id) {
+        return dispatch(updateResource(model));
+      } else {
+        return dispatch(createResource(model));
+      }
+    }
   }),
 )
 class Editor extends React.Component<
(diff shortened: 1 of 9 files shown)


src/components/editor/Editor.tsx
```diff
@@ -45,7 +45,13 @@ import {AnnotationAPICreateModel} from 'api/annotations';
   },
   dispatch => ({
     hideEditor: () => dispatch(hideEditor()),
-    createAnnotation: (data: AnnotationAPICreateModel) => dispatch(createResource(data)),
+    createAnnotation: (model: AnnotationAPICreateModel) => {
```
lgodziejewski: why `model`? Wouldn't `data` be fine? (or `annotation`)

src/components/viewer/ViewerItem.tsx
```diff
@@ -60,23 +63,18 @@ export default class ViewerItem extends React.Component<Partial<IViewerItemProps
       ViewerItem.editControlDisappearTimeout,
     );
   }
+  onEditClick = (e) => {
```
lgodziejewski: btw. in React there's a "custom" that event handlers are called `handleXY`, e.g. `onClick={handleClick}`, etc.

src/store/widgets/reducers.ts
```diff
@@ -1,22 +1,28 @@
 import {
-  EDITOR_NEW_ANNOTATION,
+  EDITOR_ANNOTATION,
   EDITOR_VISIBLE_CHANGE,
-  MENU_WIDGET_CHANGE,
+  MENU_WIDGET_CHANGE, SET_EDITOR_SELECTION_RANGE,
   VIEWER_VISIBLE_CHANGE,
 } from './actions';
-import * as _ from "lodash";
+import * as _ from 'lodash';
```
lgodziejewski: As a side note, if just a specific function is wanted (and we really care about final package size), it's better to write import as sth like `import _get from 'lodash/get'`, as then only the given fragment gets imported into our bundle.

If we just do a `import foo from 'lodash'`, then (usually) the whole package is injected into our final bundle, which often is not needed.

src/components/editor/Editor.tsx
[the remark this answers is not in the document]
kaglowka: @glowka raised the same issue :) I'm changing it in the next PR (already waiting for review). We agreed it's nice to have a universal name (used for many different *models*). We settled for `instance`, but we hadn't thought of `data` then.

src/components/viewer/ViewerItem.tsx
kaglowka: I wasn't sure about that, thanks.

src/store/widgets/reducers.ts
kaglowka: Good advice :+1:

PR #55, opened 2018-05-11
Api config

kaglowka:
- configured axios to use session cookies when connecting our API
- set prod API URL


glowka:
Does this _use_credentials_ work out of the box? :smiley:

kaglowka:
Yes, session cookies are sent in the request and the user is identified well (current server implementation assumes user = HTTP session)

PR #59, opened 2018-05-12
Editor and DraggableWidget local state management

kaglowka:
**To be merged only after #54** 

**Summary**
- meeting the goal that it should be possible to open Editor for a new annotation while one is already open. (Solution discussed in https://slack-redir.net/link?url=https%3A%2F%2Fgithub.com%2FPrzypisPowszechny%2Fpp-client%2Fpull%2F50 (outdated, beginning with "Couldn't ...") assumes Editor's opening is always preceded by setting `editor.visible` to `false`). For this reason:
    - `Editor` & `DraggableWidget` local state is managed not with constructor but with a condition in `getDerivedStateFromProps`.
    - Location is not managed by `Editor` any more, but by `DraggableWidget` now.
- component conditional rendering to App; not necessary, but as it happens, `visible` is not passed down through 3 components' props any more. (we can easily drop this change).

### DIFF
--- a/src/components/editor/Editor.tsx
+++ b/src/components/editor/Editor.tsx
@@ -50,13 +50,8 @@ class Editor extends React.Component<
   Partial<IEditorProps>,
   Partial<IEditorState>
   > {
-   /*
-    * NOTE:
-    * For a comprehensive note on invertedX and invertedY see Widget component
-    */
 
   static defaultProps = {
-    visible: true,
     locationX: 0,
     locationY: 0,
   };
@@ -74,6 +69,8 @@ class Editor extends React.Component<
      * However, it is not enough, as two annotations can be made for the exact same range
      */
     const nextAnnotation: any = nextProps.annotation || {};
+    // Note: nextProps.annotation && nextProps.annotation.id === prevState.annotationId will generate updates
+    // if only nextProps.annotation is null
     const areAnnotationsEqual = prevState.annotationId === nextAnnotation.id;
     const areRangesEqual = _.isEqual(prevState.range, nextProps.range);
     if (areAnnotationsEqual && areRangesEqual) {
@@ -88,7 +85,6 @@ class Editor extends React.Component<
         annotationLink: attrs.annotationLink || '',
         annotationLinkTitle: attrs.annotationLinkTitle || '',
 
-        moved: false,
         locationX: nextProps.locationX,
         locationY: nextProps.locationY,
         annotationLinkError: '',
@@ -104,15 +100,6 @@ class Editor extends React.Component<
     super(props);
     this.state = {};
     this.moverElement = React.createRef();
-  }
-
-  onDrag = (delta: IVec2) => {
-    this.setState({
-      locationX: this.state.locationX + delta.x,
-      locationY: this.state.locationY + delta.y,
-      moved: true,
-    });
-    return true;
   }
 
   setPriority = (priority: AnnotationPriorities) => {
@@ -228,9 +215,6 @@ class Editor extends React.Component<
 
   render() {
     const {
-      locationX,
-      locationY,
-      moved,
       priority,
       comment,
       annotationLink,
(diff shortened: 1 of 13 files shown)


src/components/widget/DraggableWidget.tsx
```diff
@@ -4,40 +4,71 @@ import styles from './Widget.scss';
...
-    return {
-      locationX: nextProps.locationX,
-      locationY: nextProps.locationY,
-      moved: false,
-    };
+  static getDerivedStateFromProps(nextProps: IDraggableWidgetProps, prevState: IDraggableWidgetState) {
+    const areInitialLocationsEqual =
+      prevState.initialLocationX === nextProps.initialLocationX
+      && prevState.initialLocationY === nextProps.initialLocationY;
+    if (areInitialLocationsEqual) {
```
glowka: This simplified equivalent (only non null returned value changes state):
```
if (!areInitialLocationsEqual) {
 return {
        initialLocationX: nextProps.initialLocationX,
        initialLocationY: nextProps.initialLocationY,
        locationX: nextProps.initialLocationX,
        locationY: nextProps.initialLocationY,
        hasBeenDragged: false,
      };
}
```

glowka: OK, that's a good point, still I would use that with one extra `return null` line :)

kaglowka: Sure, that's what I meant :)

src/containers/App.tsx
```diff
@@ -3,14 +3,32 @@ import Menu from 'components/menu';
...
+    menuVisible: state.widgets.menu.visible,
+  }),
+)
+export default class App extends React.Component<Partial<AppProps>> {
+  // always updates...:
+  // <Editor key={JSON.stringify(this.props.editor.range) + JSON.stringify(this.props.editor.annotationId)}/>
+  render() {
+    return (
+      <div>
+        {this.props.editor.visible && <Editor/>}
```
lgodziejewski: I'd propose to eventually move Widget management to some kind of `WidgetManager`, which would render appropriate widgets.

Maybe it could also provide some higher-level widget declarations, so we don't have to modify it after each new one is defined (but that's just a general future idea)

src/components/editor/Editor.tsx
```diff
@@ -75,21 +62,36 @@ class Editor extends React.Component<
...
+     * Comparing ranges is crucial when the annotation was null before and is null again;
+     * However, it is not enough, as two annotations can be made for the exact same range
+     */
+    const nextAnnotation: any = nextProps.annotation || {};
+    // Note: nextProps.annotation && nextProps.annotation.id === prevState.annotationId will generate updates
+    // if only nextProps.annotation is null
+    const areAnnotationsEqual = prevState.annotationId === nextAnnotation.id;
+    const areRangesEqual = _.isEqual(prevState.range, nextProps.range);
+    if (areAnnotationsEqual && areRangesEqual) {
+      return { ...prevState };
```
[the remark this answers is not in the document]
kaglowka: Definitely :)

src/containers/App.tsx
kaglowka: Would it involve some more widget-related logic than current `App`? I'm not sure I understand the second part. You mean not to manually insert widget components? (like `<Editor>`?)

I think it might be helpful in the future when there are more widgets involved. Right now it's quite handy to be able to manually experiment with them and not "freeze" the widget configuration.

src/components/editor/Editor.tsx
```diff
@@ -75,21 +62,36 @@ class Editor extends React.Component<
...
+    } else {
+      const attrs: Partial<AnnotationAPIModelAttrs> = nextAnnotation.attributes || {};
+      return {
+        annotationId: nextAnnotation.id,
```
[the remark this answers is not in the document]
kaglowka: Oh, and it turns out storing extra information in `state` is intentional, if we ever have any doubts, [here's a nice discussion!](https://github.com/reactjs/reactjs.org/issues/721#issuecomment-377096678)

src/containers/App.tsx
lgodziejewski: I had in mind sth like:

```
@connect(state => ({
  widgetsState: state.widgets,
})
WidgetManager {
  this.widgets = [
    Editor,
    Viewer,
    etc.,
    // add here any new ones?
  ];

  render() {
    const {
      widgetsState,
    } = this.props;

    return (
      <div>
        {this.widgets.map(widget => (
          widgetsState[widget.reducerName].visible && <widget key={widget.name} />
        )}
      </div>
    );
}
```

however, as I said, that's more of a future idea to streamline widget visibility management. As of now it probably is not really required, especially that we don't really plan to add more any more widgets (unless this could help somehow with e.g. 2+ editors being shown at once).


As of now I'd just move the rendering to a dedicated component.

PR #62, opened 2018-05-18
Ui fixes and api

glowka:


src/components/viewer/ViewerItem.tsx
```diff
@@ -61,11 +63,21 @@ export default class ViewerItem extends React.Component<Partial<IViewerItemProps
     this.setControlDisappearTimeout();
   }
 
+  componentWillMount() {
+    if (this.state.disappearTimeoutId) {
```
kaglowka: In React 16.3 this is legacy method and should theoretically be replaced with `componentDidUpdate`. However, this won't prevent the timer from firing after the component has been rendered and before `componentDidUpdate` is called? I guess no. Is this a gap in the new paradigm?

src/components/viewer/ViewerItem.tsx
```diff
@@ -61,11 +63,21 @@ export default class ViewerItem extends React.Component<Partial<IViewerItemProps
     this.setControlDisappearTimeout();
```
kaglowka: This leads to setState inside constructor (and a warning). I'll fix it in #65 .

src/store/widgets/reducers.ts
```diff
@@ -114,3 +118,21 @@ function viewerActionHandler(state, payload) {
...
+
+    if (state.viewer.annotationIds.length !== filteredAnnotationIds.length) {
+      return {
+        ...state,
+        viewer: {
+          ...state.viewer,
+          annotationIds: filteredAnnotationIds,
+          visible: filteredAnnotationIds.length > 0,
+        }};
+    }
```
kaglowka: :+1:

src/components/highlights/Highlights.ts
```diff
@@ -12,7 +12,7 @@ interface IHighlightsProps {
 
 @connect(
   state => ({
-      annotations: state.api.annotations ? state.api.annotations.data : [],
+      annotations: state.api.annotations.data
```
kaglowka: :+1:

src/components/viewer/ViewerItem.tsx
kaglowka: Or -- now that I thought of it -- `componentWillUnmount`? As far as I understand, there's no reason for the timer to be already initialized before the first mount, so it doesn't have to be checked then. On the other hand, `componentWillUnmount` covers all cases when a component is unmounted and mounted again.

What is more tricky is using an alternative to the deprecated `componentWillReceiveProps` which covers the case when the component is not actually unmounted but just receives new props. I'm wondering if there is a React16.3-ish way we could prevent the timer from before props update from firing when the component has just received new props but has not rendered yet?

src/api/annotations.ts
```diff
@@ -35,7 +39,41 @@ export interface AnnotationAPIModelAttrs {
...
+export interface Relations {
+  link: string;
+  data: APIModel[];
+}
+
+export interface APIModel {
+  id: string;
+  type: string;
+}
+
```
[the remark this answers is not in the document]
glowka: Totally agree with that. I just followed the order of classes I had found in this file (base classes to the bottom) without much consideration... which has just come by means of your words ;)

src/components/viewer/ViewerItem.tsx
glowka: Hah, using `componentWillMount` obiously does not make sense here, I think I must have wanted to use `componentWillUnmount` and this is just kind of typo (or _selecto_ as I mistakenly selected it from autocomplete  :smirk:   ) and that perfectly goes along with what you have guys proposed above!

PR #65, opened 2018-05-20
Further UI fixes

kaglowka:
**Main**
- `Viewer` does not disappear on its inner tooltip `mouseover` any more.
- TextSelector: current onSelection renamed to onMouseUp and a new, more selective event `onSelectionChange` (now the one listened to instead of onMouseUp)
- fixed ViewerItem `setState` warning.

**Changes that were a part of a rolled back solution, but are a useful extension for the future, so I decided to keep it**
- extended `TextSelector`: for selections on `.pp-ui` elements, TextSelector passes extra `isInsideArticle` argument instead of returning null on selection.

### DIFF
--- a/src/components/viewer/Viewer.tsx
+++ b/src/components/viewer/Viewer.tsx
@@ -98,12 +98,11 @@ export default class Viewer extends React.Component<Partial<IViewerProps>, Parti
   onMouseLeave = (e) => {
     // Normally, close the window, except...
     // not when the modal is not open
-    if (!this.state.confirmDeleteModalOpen) {
+    // not when this element is manually marked as an indirect Viewer child (despite not being a DOM child)
+    const isMouseOverIndirectChild = e.relatedTarget.classList.contains(PPViewerIndirectChildClass);
+    if (!this.state.confirmDeleteModalOpen && !isMouseOverIndirectChild) {
       // check what element the pointer entered;
-      // not when this element is manually marked as an indirect Viewer child (despite not being a DOM child)
-      if (!e.relatedTarget.classList.contains(PPViewerIndirectChildClass)) {
-        this.props.hideViewer();
-      }
+      this.props.hideViewer();
     }
   }
 
@@ -121,12 +120,12 @@ export default class Viewer extends React.Component<Partial<IViewerProps>, Parti
       const attrs = annotation.attributes;
       return (
         <ViewerItem
-            key={annotation.id}
-            annotation={annotation}
-            onDelete={this.onItemDelete}
-            onEdit={this.onItemEdit}
-            // ignore these elements on mouseleave
-            indirectChildClassName={PPViewerIndirectChildClass}
+          key={annotation.id}
+          annotation={annotation}
+          onDelete={this.onItemDelete}
+          onEdit={this.onItemEdit}
+          // ignore these elements on mouseleave
+          indirectChildClassName={PPViewerIndirectChildClass}
         />
       );
     });
--- a/src/components/viewer/Viewer.tsx
+++ b/src/components/viewer/Viewer.tsx
@@ -10,6 +10,7 @@ import styles from './Viewer.scss';
 import { selectViewerState } from 'store/widgets/selectors';
 import { hideViewer, showEditorAnnotation } from 'store/widgets/actions';
 import { AnnotationAPIModel } from 'api/annotations';
+import { PPViewerIndirectChildClass } from 'consts';
 
 interface IViewerProps {
   locationX: number;
@@ -95,9 +96,14 @@ export default class Viewer extends React.Component<Partial<IViewerProps>, Parti
   }
 
   onMouseLeave = (e) => {
-    // Close the window only when the modal is not open
+    // Normally, close the window, except...
+    // not when the modal is not open
     if (!this.state.confirmDeleteModalOpen) {
-      this.props.hideViewer();
+      // check what element the pointer entered;
+      // not when this element is manually marked as an indirect Viewer child (despite not being a DOM child)
+      if (!e.relatedTarget.classList.contains(PPViewerIndirectChildClass)) {
+        this.props.hideViewer();
+      }
     }
   }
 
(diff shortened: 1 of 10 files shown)


src/components/viewer/ViewerItem.tsx
```diff
@@ -206,7 +208,7 @@ export default class ViewerItem extends React.Component<Partial<IViewerItemProps
             <Popup
               trigger={this.upvoteButton()}
               size="small"
-              className="pp-ui pp-popup-small-padding"
+              className={classNames(indirectChildClassName, 'pp-ui', 'pp-popup-small-padding')}
```
lgodziejewski: can we extract these two classes to some constants file?

src/core/TextSelector.ts
```diff
@@ -131,23 +144,43 @@ export default class TextSelector {
...
-        return;
+      if (!this.isInsideArticle(container)) {
+        isInsideArticle = false;
       }
     }
 
     const serializedRanges = [];
     for (const range of selectedRanges) {
-      const serializedRange = range.serialize(this.element, '.' + PPHighlightClass);
+      const serializedRange = range.serialize(this.element, `.${PPHighlightClass}`);
```
lgodziejewski: 👍

src/init/handlers.ts
```diff
@@ -24,9 +27,14 @@ export function deinitializeCoreHandlers() {
...
-  if (selection.length === 0) {
-    store.dispatch(noSelection());
+function selectionChangeCallback(
+  selection: Range.SerializedRange[],
+  isInsideArticle: boolean,
+  event) {
+  if (selection.length === 0 || (selection.length === 1 && !isInsideArticle)) {
+    // Propagate to the store only selections fully inside the article (e.g. not belonging to any of PP components)
+    // When we need to react also to other, we can easily expand the textSelector reducer; for now it' too eager.
+    store.dispatch(makeSelection(null));
```
lgodziejewski: 👍

src/components/viewer/Viewer.tsx
```diff
@@ -115,10 +121,12 @@ export default class Viewer extends React.Component<Partial<IViewerProps>, Parti
       const attrs = annotation.attributes;
       return (
         <ViewerItem
-          key={annotation.id}
-          annotation={annotation}
-          onDelete={this.onItemDelete}
-          onEdit={this.onItemEdit}
+            key={annotation.id}
```
lgodziejewski: the key solution/"hack" is kinda OK - it has one downside however. Whenever the `annotation.id` has changed, the `ViewerItem` component/object is dropped and a new one is created.

If we have multiple annotations next to each other this can potentially introduce memory bloat.

I'd say the method `componentDidUpdate` is fine to reset timer

src/components/viewer/ViewerItem.tsx
```diff
@@ -49,36 +50,43 @@ export default class ViewerItem extends React.Component<Partial<IViewerItemProps
 
   static editControlDisappearTimeout = 500;
 
+  static getDerivedStateFromProps() {
+    return { initialView: true };
```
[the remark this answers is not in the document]
kaglowka: I thought getDerivedStateFromProps is a nice place to perform the same state initialization for mounting and updating. I'm not sure what to do about it now.

src/components/viewer/ViewerItem.tsx
```diff
@@ -49,36 +50,43 @@ export default class ViewerItem extends React.Component<Partial<IViewerItemProps
...
+
+  disappearTimeoutId: Timer;
+
   constructor(props: IViewerItemProps) {
     super(props);
-    this.state = {
-      initialView: true,
-    };
-    this.setControlDisappearTimeout();
+    this.state = {};
```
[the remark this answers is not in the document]
kaglowka: I thought so too, but when I deleted the constructor, an error was thrown that the state has not been initialized... unless I'm missing something.

src/components/viewer/ViewerItem.tsx
kaglowka: I'd love to. I didn't want to mix it with other changes (I guess `pp-ui` appears in some 15 places now). It's coming in the next PR ;)

src/components/viewer/Viewer.tsx
lgodziejewski: Nvm, I haven't noticed that it's generated from an array

src/components/viewer/ViewerItem.tsx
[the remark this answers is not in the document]
lgodziejewski: "the constructor" ?

src/components/viewer/Viewer.tsx
kaglowka: Just to document our Slack discussion:
We're not sure what is the chance that the timer set with previous props/annotation is going to fire before it is cleared in `componentDidUpdate`.

According to [this post from StackOverflow](https://stackoverflow.com/questions/33613728/what-happens-when-using-this-setstate-multiple-times-in-react-component) setState calls in result to setTimeout are not batched, so we can't be sure if clearing timer as late as in `componentDidUpdate` is safe.

Since we're listing annotations, we need to set `key` anyway, but it's interesting how to deal with with single child cases, when setting `key` is not necessary. It seems so far, that even in such case it is the only foolproof solution we know of.

src/components/viewer/ViewerItem.tsx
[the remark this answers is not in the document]
lgodziejewski: a potential side-effect of using `getDerivedStateFromProps` in this scenario:
`Note that if a parent component causes your component to re-render, this method will be called even if props have not changed. You may want to compare new and previous values if you only want to handle changes.`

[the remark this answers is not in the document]
kaglowka: One more thing to consider - `setState` in `componentDidUpdate` will surely cause the component to rerender? I'd say this is quite undesirable as it adds complexity.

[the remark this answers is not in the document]
lgodziejewski: Correct, my suggestion makes no sense taking that into consideration 😛

Issue #74, opened 2018-05-25
Final PP 0.1 refactor


kaglowka:
- [x] use `requestAnimationFrame` (#49 )
- [x] replace React Highlights component with pure Redux (#50)

PR #77, opened 2018-05-27
74 replace timeout with raf rework highlights

lgodziejewski:
Implements #74


src/init/highlights.ts
```diff
@@ -0,0 +1,54 @@
...
+  instance.annotations = annotations;
+  // and redraw
+  instance.highlighter.drawAll(annotations.map(annotation => ({
+    id: annotation.id,
+    range: annotation.attributes.range,
+    annotationData: annotation,
+  })));
+}
+
+function handleHighlightEvent(e, annotations) {
```
kaglowka: I'd say this handler name could be more semantic (handleHighlightMouseOver) as it is attached precisely to `mouseover` event.

src/init/highlights.ts
```diff
@@ -0,0 +1,54 @@
+import store from 'store';
+import { showViewer } from 'store/actions';
+import { mousePosition } from 'common/dom';
+import Highlighter from 'core/Highlighter';
+
+let instance;
```
kaglowka: Why not just declare a ES6 class? Since it would be pretty much equivalent, I'd personally choose a class for readability. Do you find this pattern you're using more common/natural?

lgodziejewski: This is a "flattened" (due to es6 modules(files)) version of the JS "module pattern", so it's a rather common one.

kaglowka:
I'm afraid "implements" does not close an issue? By the way, neither does a sequence of issues with a single keyword like I once guessed. (https://help.github.com/articles/closing-issues-using-keywords/).

lgodziejewski:
I didn't even know about these keywords 😛 Will try to use `resolves` next time

kaglowka:
Haha, at first I thought `implements` was a new keyword I didn't know about and I was happy they introduced it, but... :)

Issue #95, opened 2018-06-10
"add annotation" mode


kaglowka:

KasiaSkowronska:
add annotation user flow:
https://drive.google.com/open?id=12v4IXZTKRJEc2mrVW9ebbJgQp0GG9FiG

"Poproś o przypis" should be disabled for now

Issue #94, opened 2018-06-10
Annotation display switch in browser popup.


kaglowka:

glowka:
Is "popup" official name for that part of extension or is just our own term?
I am asking because it takes me some effort to associate this kind of drop-down menu with popup, but I am not going to argue with with some broader trends...

kaglowka:
Yes, `popup` is an official name for that part of extension...

PR #108, opened 2018-06-16
Browser-extension popup interface + basic app modes interaction

KasiaSkowronska:
new popup interface, related to #94 and #95 

- KG, I'm sorry, I've made our own toggle, I just hate the idea that they could be in wrong order. I hope it's ok. I guess this toggle folder should be somewhere else.

- I hope you can fix this error with loading icons


src/browser-extension/popup/toggle/toggle.scss
```diff
@@ -0,0 +1,56 @@
...
+  background-color: #2ecc71;
+}
+
+input:focus + .slider {
+  box-shadow: 0 0 1px #2ecc71;
+}
+
+input:checked + .slider:before {
+  -webkit-transform: translateX(15px);
+  -ms-transform: translateX(15px);
```
kaglowka: We don't need to be compatible with other browsers right now, and when we need, we can have [autoprefixer](https://github.com/postcss/autoprefixer) do it for us ;)

src/browser-extension/popup/toggle/toggle.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+
+import './toggle.scss';
+
+declare const chrome: any;
+
+export default class Toggle extends React.Component<{}, {}> {
+  constructor(props: {}) {
+    super(props);
```
kaglowka: I think it's necessary for this component to contain text that is clickable and allows to activate/deactivate the toggle button.
I'll fix it soon.

kaglowka:
TODO
- [ ] not display annotations on the page startup.
- [ ] fix "disable on this page" mode.

kaglowka:
Replaced by #128

PR #110, opened 2018-06-18
PP 0.2 editor validation

glowka:


src/components/editor/Editor.tsx
```diff
@@ -135,6 +137,11 @@ class Editor extends React.Component<
     if (!this.state.annotationLinkTitle) {
       this.setState({ annotationLinkTitleError: 'Musisz podać tytuł źródła, jeśli chcesz dodać przypis!' });
       return false;
+    } else if (this.state.annotationLinkTitle.length > Editor.linkTitleMaxLength) {
+      this.setState({ annotationLinkTitleError:
+          `Skróć tytuł źródła z ${this.state.annotationLinkTitle.length} do ${Editor.linkTitleMaxLength} znaków!`,
```
kaglowka: I guess with the current function size it might be a little bit more elegant with something like
```js
const { 
  annotationLinkTitle, 
  annotationLinkTitleError,
  (...)
} = this.state;
```
But that's just a thought.

src/components/editor/Editor.tsx
```diff
@@ -133,6 +135,14 @@ class Editor extends React.Component<
     if (!this.state.annotationLink) {
       this.setState({ annotationLinkError: 'Musisz podać źródło, jeśli chcesz dodać przypis!' });
       return false;
+    } else if (this.state.annotationLink.length > Editor.linkMaxLength) {
+      this.setState({ annotationLinkError:
+          `Skróć źródło z ${this.state.annotationLink.length} do ${Editor.linkMaxLength} znaków!`,
+      });
+      return false;
+    } else if (!isValidUrl(this.state.annotationLink)) {
```
kaglowka: If `isValidUrl` expected protocol presence and if we explicitly normalized it, we could achieve clear separation into "normalization" and "validation" stage (where `isValidUrl` "takes no hostages" -- doesn't assume the https-missing link will be corrected before dispatching it to the server)
We would than validate it as follows:
```js
isValidUrl(httpPrefixed(this.state.annotationLink))
```

src/utils/url.js
```diff
@@ -1,18 +1,32 @@
...
+export function httpPrefixed(url) {
+  if (url.indexOf("http://") === 0 || url.indexOf("https://") === 0 ) {
+    return url;
+  }
+  return "http://" + url;
+}
 
-    return hostname;
+export function isValidUrl(url) {
+  return url.match(urlRegex)
```
kaglowka: :+1:

src/components/editor/Editor.tsx
glowka: Just a thought, but good one :P

glowka:
Hey, one general thought. If you have some comments that are not blocking merging, but suggest some changes, it would be nice not to merge the PR, but leave it to the assignee to decide whether to A) append the changes before merging or B) merge it at once and introduce the changes later on.

I know as for now we used the merging as the sign of approval for changes as a whole, but in some situation it is event nicer to get positive feedback, but then make it even better and merge one "handy" PR with all changes at once.

@kaglowka Are you OK with that idea?

kaglowka:
Yeah :) sure, I thought you might prefer to have it merged, too, and correct it later on if you wish. But once it's clear it's up to the assignee to merge it or correct it before that, I'll gladly just set the the approval.

PR #114, opened 2018-06-20
ReportEditor

glowka:


src/components/viewer/report-editor/Report.tsx
```diff
@@ -0,0 +1,122 @@
...
+  handleInputChange = (e: React.ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => {
+    const target = e.currentTarget;
+    const stateUpdate = {
+      [target.name]: target.value,
+      showReasonError: target.name !== 'reason' && this.state.showReasonError,
+    };
+    this.setState(stateUpdate);
+  }
+
+  getRadioId(reason: Reasons) {
```
kaglowka: Do we need `id` for anything?

src/components/viewer/report-editor/Report.tsx
```diff
@@ -0,0 +1,122 @@
...
+    }
+    this.props.onSubmit(this.state.reason, this.state.comment);
+  }
+
+  render() {
+    return (
+      <div className={classNames(PPScopeClass, styles.self, styles.editor)}>
+        <h3>Co jest nie tak?</h3>
+        <div>
+          <div className={classNames(styles.radioInputLine)}>
```
kaglowka: I think this could be written more concisely with something like
```js
[Reasons.BIASED, (...)].map( reason => (
    <div (...) >
         <input
         value={reason}
         (...)
         >
</div>
));
```

Just to economize on typing, of course. Since for each such construction there is a time, when some radio button needs an individual change and you need to expand it... but I think it's nice to start off.

src/components/viewer/report-editor/ReportEditor.scss
```diff
@@ -0,0 +1,137 @@
...
+  background-image: url('../../../../assets/report-symbol.svg');
+}
+
+.suggest-icon::after {
+  @extend .icon;
+  background-image: url('../../../../assets/suggest-edit-symbol.svg');
+}
+
+.hide {
+  display: none !important;
```
kaglowka: This will probably stop working, when we use `css-important-loader` for PP 0.2. 

Maybe we should apply a nesting trick like `.self.hide`?
In some cases we can also perhaps stop displaying the element altogether in React rather than apply CSS classes (we probably can't e.g. for animated transitions. Are there any more problematic cases?)

src/components/viewer/report-editor/ReportEditor.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+interface IReportEditorProps {
+  annotation: AnnotationAPIModel;
+  onCancel: (e) => void;
+  onSuccess: () => void;
+  createAnnotationReport: (instance: AnnotationReportAPICreateModel) => Promise<AnnotationReportAPIModel>;
+}
+
+interface IReportEditorState {
+  activeDialog: 'menu'|'report'|'suggestion';
```
kaglowka: Why not use the `enum Dialogs` since it's already declared and used?

src/components/viewer/report-editor/ReportEditor.tsx
```diff
@@ -0,0 +1,113 @@
...
+              </button>
+            </div>
+          </div>
+        );
+      case 'report':
+        return <Report annotation={annotation} onCancel={onCancel} onSubmit={this.save}/>;
+      case 'suggestion':
+        return <Suggestion annotation={annotation} onCancel={onCancel} onSubmit={this.save}/>;
+      default:
+        return;
```
kaglowka: When fixing the error for Chrome, I discovered that `return;` caused React to crash.
`return null` on the other hand will always keep the application going, so how about always using `return null` in similar cases?

src/components/viewer/report-editor/ReportEditor.tsx
```diff
@@ -0,0 +1,113 @@
...
+                <span> Zasugeruj poprawkę </span>
+              </button>
+            </div>
+          </div>
+        );
+      case 'report':
+        return <Report annotation={annotation} onCancel={onCancel} onSubmit={this.save}/>;
+      case 'suggestion':
+        return <Suggestion annotation={annotation} onCancel={onCancel} onSubmit={this.save}/>;
+      default:
```
kaglowka: I like the naming distinction between `onSubmit` and `onSuccess` based on whether the data is dispatched to the server and there is room for failure -- I think it's a nice distinction to always keep.

src/css/common/mixins.scss
```diff
@@ -3,3 +3,20 @@
...
+  padding: 5px;
+  font-size: $default-font-size + 2px;
+
+  i {
+    margin: 0;
+  }
+
+  &[disabled] {
+    display: none;
+  }
```
kaglowka: :+1:

src/components/viewer/report-editor/ReportEditor.scss
glowka: I am not sure if `!important` is at all necessary here, I just copied existing solution, so it's broader issue which we could deal with as a part of task `css-imporant-loader`: *check all `!important` cases and apply better specificity(https://www.w3.org/TR/selectors-3/#specificity) if needed*

src/components/viewer/report-editor/ReportEditor.tsx
glowka: ups, stale leftover, I'll fix it, thx

src/components/viewer/report-editor/ReportEditor.tsx
glowka: Before passing the chrome issue to you I spotted this return and replaced it with return null and nothing changed, so I either we had two bugs (one that you fixed by your commit and this one) or this is not the prettiest code but does not lead to crashes.

src/components/viewer/report-editor/Report.tsx
glowka: To labels pair with inputs, so that you can check the input by clicking on the label.

src/components/viewer/report-editor/Report.tsx
glowka: Thanks for idea, b for each input there is unique label, so this case is bit more complicated and I would leave it as it is. Anyway, it's also just a couple of static inputs, so I am not sure if any solution is better or worse here :yum:

glowka:
Thanks for the fix.
I understand that you develop on chrome and to release plugin on chrome you need to test it on chrome, not on anything else. However as long as I do tasks that are just pure JS+React it would be nice to write good code that is JS not *chromeJS*. I work, do *all* the stuff on firefox, so as long as PP tasks are feasable there I would like to do it that way.
If any issue come up I can fix it as part of task, unless we find the cause of the bug to be somewhere far far from the task, which we had thought to be the case this time. It surprisingly turned out to be much closer - in my code, so... thanks again for the fix :sweat_smile:

Issue #127, opened 2018-06-25
Remove Menu


kaglowka:
Remove Menu in annotation mode as outlined in #95 

@KasiaSkowronska : please confirm or close this issue otherwise.

kaglowka:
I assume we're not removing the menu. If not true, reopen the issue @KasiaSkowronska

Issue #126, opened 2018-06-25
Finish the browser popup


kaglowka:
After  #108, we need to refine the 

- [ ] grayed out icon(s)
- [ ] what to do with the "poproś o przypis" button?
- [ ] what to do with "daj znać co myślisz" button?

KasiaSkowronska:
I suggest we add tooltip to "poproś o przypis" and "daj znać co myślisz" buttons with something like "Opcja jeszcze niedostępna w tej wersji" and make them disabled. If you would like to know why I think it's better solution then removing those buttons, we can talk (but also I think it's not a big deal and it may be a waste of time to discuss it).

I don't understand what is this "grayed out icon(s)" task

kaglowka:
Does it make sense to include a button with information it's not available in this version?

kaglowka:
Secondly, I've never seen an inactive button with a tooltip on mouse hover. Maybe just gray them out with a small static label like "Niedostępne w tej wersji".

KasiaSkowronska:
ok, just remove them

kaglowka:
Ok. Never mind the "grayed out" point. It's not necessary.

Issue #124, opened 2018-06-24
Annotation mode and disabled extension mode interaction


kaglowka:
A bunch of basic questions.
- should it be possible to turn on the annotation mode when in the disabled extension mode?
- should an already open Editor disappear when the extension is disabled?
- should an already open Editor disappear when the annotation mode is turned off?

Add your own if anything more comes to your mind.

KasiaSkowronska:
- no
- yes
- yes

Also I think it would be nice to handle this scenario when user first highlights text and then clicks "dodaj przypis" -- in that case the Editor should open. Now it doesn't work like that and user has to highlight text twice.

kaglowka:
I opened an issue for the scenario you mentioned. #142

PR #128, opened 2018-06-25
Browser-extension popup interface + app modes

### LINKED ISSUE #94: Annotation display switch in browser popup.

### LINKED ISSUE #95: "add annotation" mode

kaglowka:
resolves #94 
resolves #95

I'm sorry, this PR has grown much beyond the size comfortable to review. A clear distinction could have been done e.g. between the interface and the business logic and let's consider it a recommendation for the future.

**summary**
- new browser popup interface
- features that can be turned on in the popup:
    - "add annotation" mode
    - "turn off the extension"
    - "turn off the extension for this page"
- an annotation mode widget visible in the annotation mode
- chrome storage used for communication between browser popup and content scripts
- temporarily, for "client" configuration, chrome storage is mocked for this configuration not to stop working altogether; this is a quick temporary solution, that we should work on or drop soon. For now, annotation mode is always enabled in client configuration.
- From [two available kind of Chrome storages](https://developer.chrome.com/apps/storage), `local` was chosen for development and `sync` for production due to [quota on sync usage per browser extension](https://developer.chrome.com/extensions/storage#type-StorageArea)

**Several changes still need to be done or considered**
#124
#126
#127


src/index.tsx
```diff
@@ -34,9 +39,44 @@ function loadInitialData() {
   store.dispatch(readEndpoint('/annotations?url=' + window.location.href));
 }
 
+export function loadDataFromChromeStorage() {
```
glowka: All other actions taking place here are just one-liners and this function take most of the file, is it that important? How about moving it to less prominent place?

src/store/appModes/reducers.ts
```diff
@@ -0,0 +1,29 @@
...
+const initialState: AppModeReducer = {
+  isExtensionDisabled: false,
+  annotationModePages: [],
+  disabledPages: [],
+};
+
+export default function appModes(state = initialState, action) {
+  switch (action.type) {
+    case MODIFY_APP_MODES:
+      return appModesActionHandler(state, action.payload);
```
glowka: Calling this trivial function does not make it more descriptive to me and it definitely gets harder to read with every action being trivial function that I have to follow just to find it's trivial. Do we need this layer? Why not using `{ ...state,  ...payload,  }` directly?

src/store/reducer.ts
```diff
@@ -1,14 +1,16 @@
 import { combineReducers } from 'redux';
 import { reducer as api } from 'redux-json-api';
 import widgets, { WidgetReducer } from './widgets/reducers';
+import appModes, { AppModeReducer } from './appModes/reducers';
 import textSelector from './textSelector/reducers';
 import { AnnotationAPIModel, AnnotationUpvoteAPIModel } from 'api/annotations';
 
-export interface IStore {
+export interface IAppState {
```
glowka: Nice change but maybe go even further and be even more descriptive like `IPageState` or `ITabState`?

src/browser-extension/popup/BrowserPopup.tsx
```diff
@@ -1,41 +1,180 @@
...
+
+    this.loadStateFromStorage();
+  }
+
+  componentDidUpdate() {
+    this.loadStateFromStorage();
+  }
+
+  isExtensionDisabledForCurrentTab() {
+    return (this.state.disabledPages || []).indexOf(this.state.currentTabUrl) !== -1;
```
lgodziejewski: I think using [includes](https://devdocs.io/javascript/global_objects/array/includes) (as it is finally available) is a bit more readable

src/browser-extension/popup/BrowserPopup.tsx
```diff
@@ -1,41 +1,180 @@
...
+      currentTabUrl,
+    } = this.state;
+
+    const isAnnotationMode = this.isAnnotationModeForCurrentTab();
+    if (!isAnnotationMode) {
+      let newAnnotationModePages;
+      newAnnotationModePages = [...annotationModePages, currentTabUrl];
+      this.setState({ annotationModePages: newAnnotationModePages });
+      chromeStorage.set({ [chromeKeys.ANNOTATION_MODE_PAGES]: newAnnotationModePages });
+      window.close();
```
lgodziejewski: doesn't this close the tab?

src/browser-extension/popup/BrowserPopup.tsx
```diff
@@ -1,41 +1,180 @@
...
+    }
+  }
+
+  handleDisabledExtensionChange = (e: ChangeEvent<HTMLInputElement>) => {
+    const newValue = e.target.checked;
+    this.setState({ isExtensionDisabled: newValue });
+    chromeStorage.set({ [chromeKeys.DISABLED_EXTENSION]: newValue });
```
lgodziejewski: as the `setState` and `chromeStorage.set` pair is repeated here and there, maybe it'd be good to wrap them in a function (e.g. `setPersistedState`)?

src/browser-extension/popup/toggle/toggle.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+import './toggle.scss';
+
+declare const chrome: any;
```
lgodziejewski: is it needed here?

src/components/editor/Editor.tsx
```diff
@@ -294,7 +299,7 @@ class Editor extends React.Component<
           className={styles.close}
           onClick={this.onCancelClick}
         >
-          <i className="remove icon" />
+          <i className="remove icon"/>
```
lgodziejewski: Maybe we should configure tslint to define a style for stuff like this (and below)? (unless it already is configured and I'm not aware of it 😛)

src/index.tsx
```diff
@@ -34,9 +39,44 @@ function loadInitialData() {
...
+  return new Promise((resolve, reject) => {
+    chromeStorage.get([
+      chromeKeys.ANNOTATION_MODE_PAGES,
+      chromeKeys.DISABLED_EXTENSION,
+      chromeKeys.DISABLED_PAGES,
+    ], (result) => {
+      console.log(result);
```
lgodziejewski: .

src/index.tsx
lgodziejewski: 👍, I'd say it should be moved to some utils/actions/ file

src/init/chromeStorageHandlers.ts
```diff
@@ -0,0 +1,22 @@
...
+
+const storageKeysToAppMode = {
+  [chromeKeys.DISABLED_EXTENSION]: 'isExtensionDisabled',
+  [chromeKeys.ANNOTATION_MODE_PAGES]: 'annotationModePages',
+  [chromeKeys.DISABLED_PAGES]: 'disabledPages',
+};
+
+export default function initializeChromeStorageHandlers() {
+  chromeStorage.onChanged.addListener((changes, namespace) => {
+    const newModes = {};
```
lgodziejewski: unused?

src/init/chromeStorageHandlers.ts
```diff
@@ -0,0 +1,22 @@
...
+    for (const key of Object.keys(changes)) {
+      store.dispatch(
+        changeAppModes({ [storageKeysToAppMode[key]]: changes[key].newValue } as AppModeReducer),
```
lgodziejewski: maybe it'd be better to construct `newModes` as
```
newModes = Object.keys(changes).reduce(acc, key => acc[key] = changes[key].newValue, {})
```
and then use a single `store.dispatch` call with the whole object?

Now there is a separate dispatch for every `key` available in `changes`

src/store/appModes/reducers.ts
lgodziejewski: Usually action handlers are split into functions if they are a bit more complex (and if there's a lot of them), but in this case I'd say we can do it like @glowka suggested, especially that it's rather very simple.

Unless we want to have a strict style of writing reducers (maybe better for general code readability - same solutions used everywhere, one decision less to make when writing new code)

src/store/appModes/selectors.ts
```diff
@@ -0,0 +1,20 @@
...
+  state => state.appModes,
+  (appModes) => {
+    // Standardize the URL by disregarding stuff that does not identify a page like URL parameters etc.
+    const currentStandardizedUrl = standardizeUrlForPageSettings(window.location.href);
+
+    return {
+    // Page highlights are disabled when
+    //  the extension is disabled in general or when it is disabled for this particular page
+      arePageHighlightsDisabled: appModes.isExtensionDisabled
+      || appModes.disabledPages.indexOf(currentStandardizedUrl) !== -1,
```
lgodziejewski: `includes`

src/utils/url.js
```diff
@@ -30,3 +29,8 @@ export function httpPrefixed(url) {
 export function isValidUrl(url) {
   return url.match(urlRegex)
 }
+
+export function standardizeUrlForPageSettings(url) {
```
lgodziejewski: There is a nice npm package for url operations like this - https://www.npmjs.com/package/url-parse (and some of the functions above)

kaglowka:
> however please keep it shorter in the future

Couldn't agree more :)

src/browser-extension/popup/BrowserPopup.tsx
```diff
@@ -1,41 +1,180 @@
...
+          isExtensionDisabled: result[chromeKeys.DISABLED_EXTENSION] || false,
+          disabledPages: result[chromeKeys.DISABLED_PAGES] || [],
+        });
+      });
+    }
+
+  }
+
+  componentDidMount() {
+    chrome.tabs.query({ active: true, currentWindow: true }, (tabs) => {
```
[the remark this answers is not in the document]
kaglowka: I completely agree -- I didn't want to come into the browser-independent overhead just yet.

src/browser-extension/popup/BrowserPopup.tsx
kaglowka: Thanks, didn't know about it :+1:

src/browser-extension/popup/BrowserPopup.tsx
kaglowka: BrowserPopup lives in a completely separate environment, and `window` is in this case the browser popup area.

src/init/chromeStorageHandlers.ts
kaglowka: :+1:

src/utils/url.js
kaglowka: Thanks, it's good to know.

PR #133, opened 2018-06-29
PP 0.2 clean-up after ReportEditor and other works

### LINKED ISSUE #119: Refactor after ReportEditor works
- [x] use shorter reference to `this.state.x` in `Editor`
- [x] split `api/annotations` models
- [x] move props and state declaration from `components/editor/interfaces` back to `components/editor/Editor`

glowka:
Resolves #119


glowka:
After inital commit *Split api/annotations*  a lot of misleading webpack warnings appeared.

There is nasty typescript issue with exporting interfaces, it shows up, when interfaces are not directly imported to be used, but to be further re-exported, which is necessary for creating nice index.tsx entry point for importing names from whole directory simply using 'api'.
This issue is a completely fatal architecure-based wontfix, which messed a lot in typescript+webpack world: https://github.com/webpack/webpack/issues/2977#issuecomment-245887841

The latest commit abandons idea of nice importing from single 'api' location, instead imports must be separate eg 'api/annotations', 'api/annotations-reports' etc. Last commit from #137 is alternative implementation that replaces interfaces with abstract classes, but saves index single 'api' directory import. Which approach suits you best @kaglowka @lgodziejewski?

src/api/json-api.ts
```diff
@@ -0,0 +1,23 @@
+export interface Relation {
+  link?: string;
+  data: APIModel | null;
+}
+
+export interface CreateRelation {
+  data: APIModel | null;
+}
+
+export interface Relations {
```
kaglowka: Maybe something like `oneToManyRelation`?
It's longer, but I find this distinction a bit enigmatic ;)

kaglowka:
Regarding the import issue you are mentioning:

I find using abstract classes (instead of interfaces) unncessary for two reasons (mostly the first):
1) they can be misleading -- "why are we using an abstract class instead of an interface here? Will this abstract class be overriden by a class that is implemented?"
2) they add volume to the generated code (slightly heavier and slowe packages)

Finally, a subtle issue -- again for clarity, I don't think we should use abstract classes unless we really find a use case for them (a logically legitimate use case). Most of us are only just learning Typescript, and as far as I know abstract classes weren't even included in the initial Typescript releases.

It's not that important for me to use a shortcut to the imported interface, so I'm voting for the original option in this PR ;)

PR #136, opened 2018-06-30
Transparent offset around widget

### LINKED ISSUE #134: Offset around viewer to prevent it from hiding on accidental mouse leave
By adding transparent div that is part of the viewer, so that mouse leave event is not triggered.

glowka:


src/components/widget/utils.ts
```diff
@@ -3,6 +3,8 @@ export function isInverted(widget, window) {
     // console.log('orientation');
     // TODO replace annotator jQuery with pure js
     const win = $(window);
+    // This number must be synchronized with $transparent-offset from Widget.scss
+    const transparentOffset = 50;
```
kaglowka: I think it could be nice to keep such constants in a separate file.
`class_consts.ts` function was meant to control the classes used both in JS and CSS.

The name gets a bit unfortunate, but just because I didn't think it should be necessary to share anything else than classes between CSS and JS;
I'm just communicating the idea behind that file, and I think it would be nice to extend it.

glowka: Fixed - I found simple way to export this const from css module

PR #137, opened 2018-06-30
WIP: PP 0.2 cleanup - last commit applying abstract classes to fix webpack interface related warnings

glowka:


glowka:
Decided to use alternative version #133

PR #140, opened 2018-07-02
Pp02 css modules export

glowka:


src/class_consts.ts
```diff
@@ -1,12 +1,13 @@
...
 
 /*
  * Elements wearing this class will be ignored in viewer mouseleave handling
  * (their mouseenter will not trigger viewer disappearance)
  */
 export const PPViewerIndirectChildClass = 'pp-viewer-indirect-child';
 
-// IMPORTANT: Keep it in sync with css/common/vars/scope
-export const PPScopeClass = 'pp-ui';
+export const PPScopeClass = scopeStyles['pp-scope-class'];
```
kaglowka: This is quite hot :)
Not such an obvious feature of CSS modules.

src/components/viewer/report-editor/ReportEditor.scss
```diff
@@ -58,14 +58,15 @@ $menu-inner-width: 165px;
...
-  // override semantic-ui
-  :global(.error-msg.ui.pointing.red.basic.label.large) {
-      font-size: 14px;
-      position: absolute;
-      z-index: $viewer-z-index + 2;
-      margin-top:0.5em;
-  }
+
+// override semantic-ui
+:local(.error-msg):global(.ui.pointing.red.basic.label.large) {
```
kaglowka: So, by the look of it, if you use `:global` within a selector, you need to explicitly use `:local` or all classes are going to be global?

glowka: I thought so, but after a while I am not sure. Last time it had behaved just like you said, but then I additionally used a bit more complicated expression: 
```
.error-msg {
  &:global(.ui.pointing...) {
    ...
  }
}
```
So now I think that might have been SASS that was responsible for making `:global` selector containing all the classes, I need to check that, because the option you suggest is still possible.

lgodziejewski: In CSS Modules-loaded file, `:local` is used by default

https://github.com/webpack-contrib/css-loader#modules

glowka: Yes, that's where I learn about those selectors, but as I wrote above it didn't work as expected when combined with a bit more complicated declaration. But if it was SASS responsible for that, everything would work like you said and `:local` is redundant.

PR #141, opened 2018-07-03
Basic context menu annotation option

### LINKED ISSUE #135: Right-mouse-button way of adding annotation

kaglowka:
Resolves #135 

For simplicity, context menu way of opening Editor involves no check the annotation is made within the article content. (#143)

The mechanism:
- context menu "dodaj przypis" click is detected by the background script
- backgrond script posts a message to the currently active tab via browser API
- the active tab's content script receives the message and opens the editor for the current selection

### DIFF
--- a/config/browser-extension/ext.base.config.js
+++ b/config/browser-extension/ext.base.config.js
@@ -23,6 +23,7 @@ const manifest = merge(manifestSettings.base, {
     default_popup: 'popup.html',
   },
   icons: {
+    // This size is used in context menu
     '16': 'assets/icon.png',
   }
 });
--- a/config/browser-extension/ext.base.config.js
+++ b/config/browser-extension/ext.base.config.js
@@ -22,6 +22,9 @@ const manifest = merge(manifestSettings.base, {
     default_icon: 'assets/icon.png',
     default_popup: 'popup.html',
   },
+  icons: {
+    '16': 'assets/icon.png',
+  }
 });
 
 const config = (env, argv) => merge(common.config(env, argv), {
--- a/src/core/TextSelector.ts
+++ b/src/core/TextSelector.ts
@@ -13,6 +13,7 @@ const TEXTSELECTOR_NS = 'pp-textselector';
  *
  * It is PP's adaptation of annotator's TextSelectors
  */
+
 /*
  * IMPORTANT NOTE on SELECTION RANGES
  * We leave annotator's TextSelector more or less as it is;
@@ -129,6 +130,24 @@ export default class TextSelector {
     }
 
     return ranges;
+  }
+
+  currentSingleSelectionCenter = () => {
+    /*
+     * We assume only a single selection is made
+     * Return null if zero or more selection ranges are made
+     */
+    const selection = window.getSelection();
+    if (selection.rangeCount !== 1) {
+      return null;
+    }
+    const selectionBox = selection.getRangeAt(0).getBoundingClientRect();
+    const x = window.scrollX + selectionBox.left + selectionBox.width / 2;
+    const y = window.scrollY + selectionBox.top + selectionBox.height / 2;
+    return {
+      x,
+      y,
+    };
   }
 
   /**
--- a/src/core/TextSelector.ts
+++ b/src/core/TextSelector.ts
@@ -149,7 +149,7 @@ export default class TextSelector {
   currentSingleSelectionCenter = () => {
     /*
      * We assume only a single selection is made
-     * Return null if zero or more selection ranges are made
+     * Return null if less or more than one selection range is made
      */
     const selection = window.getSelection();
     if (selection.rangeCount !== 1) {
--- a/src/core/TextSelector.ts
+++ b/src/core/TextSelector.ts
@@ -132,6 +132,20 @@ export default class TextSelector {
     return ranges;
   }
 
+  currentSerializedSelection = () => {
+    const selectedRanges = this.captureDocumentSelection();
+    return this.serializeRanges(selectedRanges);
+  }
+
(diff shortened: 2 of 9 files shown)


src/init/documentHandlers.ts
```diff
@@ -51,3 +54,22 @@ function selectionChangeCallback(
...
+     * For now, do not check for being inside article.
+     * Reason: checking ContextMenu API selection for being insideArticle is possible, but uncomfortable,
+     * as context menu actions are handled in the separate background script.
+     */
+    const selection = handlers.selector.captureDocumentSelection();
+
+    if (selection.length === 1) {
+      store.dispatch(setSelectionRange(selection[0]));
+      const selectionCenter = handlers.selector.currentSingleSelectionCenter();
+      store.dispatch(showEditorAnnotation(selectionCenter.x, selectionCenter.y));
```
glowka: Does it effectively mean that the editor appears in more or less the same place as it happens if it is invoked _normally_ (using "add annotation") mode?

kaglowka: As it happens (rather to my surprise), the window selection's bounding box can be straightforwardly accessed.
From context menu, Editor opens at the center of the selection. I found it good enough. For now, in annotation mode it opens at the mouse location.
Yes, it's slightly inconsistent.
Which way is better? I think both are good enough and the inconsistency itself doesn't compromise the usability. So I think making it consistent is a matter for another sprint.

src/core/TextSelector.ts
```diff
@@ -157,11 +190,7 @@ export default class TextSelector {
       }
     }
 
-    const serializedRanges = [];
-    for (const range of selectedRanges) {
-      const serializedRange = range.serialize(this.element, `.${PPHighlightClass}`);
-      serializedRanges.push(serializedRange);
-    }
+    const serializedRanges = this.serializeRanges(selectedRanges);
```
lgodziejewski: 👍

src/core/TextSelector.ts
```diff
@@ -131,6 +132,38 @@ export default class TextSelector {
     return ranges;
   }
 
+  currentSerializedSelection = () => {
+    const selectedRanges = this.captureDocumentSelection();
+    return this.serializeRanges(selectedRanges);
+  }
+
+  serializeRanges = (ranges: Range.NormalizedRange[]) => {
```
lgodziejewski: if you'd like this function to be super slick (😛), you could write it like:
```
return ranges.reduce((acc, range) => acc.push(range.serialize(this.element, `.${PPHighlightClass}`)), []);
```

or even better:
```
ranges.map(range => range.serialize(this.element, `.${PPHighlighClass}`);
```

lgodziejewski: Actually, apart from one-liner tricks, these functions can have a bit better performance. 

Assuming a large collection in `ranges`, the `for` loop actually blocks the js thread for its whole duration (possibly freezing the UI).

`reduce` and `map` on the other hand are callback-based, so each call(or chunks of them) are done in separate events in the event loop, allowing the UI to work normally in the meantime

lgodziejewski: On the other hand - I guess `selectedRanges` (line 193) is never going to be large? IIRC it's usually 1 element?

src/init/documentHandlers.ts
```diff
@@ -25,10 +26,12 @@ export function initializeDocumentHandlers() {
   };
 
   highlights.init(highlighter);
+  chrome.runtime.onMessage.addListener(contextMenuAnnotateCallback);
```
lgodziejewski: maybe it'd worth to hide it behind an abstraction?

config/browser-extension/ext.base.config.js
```diff
@@ -22,6 +22,9 @@ const manifest = merge(manifestSettings.base, {
     default_icon: 'assets/icon.png',
     default_popup: 'popup.html',
   },
+  icons: {
+    '16': 'assets/icon.png',
```
lgodziejewski: why '16'? 😛

src/init/documentHandlers.ts
```diff
@@ -51,3 +54,22 @@ function selectionChangeCallback(
...
+
+function contextMenuAnnotateCallback(request, sender) {
+  if (request.action === 'ANNOTATE') {
+    /*
+     * For now, do not check for being inside article.
+     * Reason: checking ContextMenu API selection for being insideArticle is possible, but uncomfortable,
+     * as context menu actions are handled in the separate background script.
+     */
+    const selection = handlers.selector.currentSerializedSelection();
+    console.log(selection);
```
lgodziejewski: .

src/init/documentHandlers.ts
glowka: So more or less the same behavior - I am perfectly happy with that :smile:

src/core/TextSelector.ts
kaglowka: Oh, I didn't know about this blocking/non-blocking difference. It's good to know.

But yes, unforutantely it's true that we more or less asume there won't be more than 1 annotation range :)

src/init/documentHandlers.ts
kaglowka: I reflected it might be good to keep the code "lean" sometimes and expose these details -- especially for people who are not very accustomed to browser API (which is everybody here except me at the very moment of writing it)
But yes, I agree it's worth changing at the first opportunity

config/browser-extension/ext.base.config.js
kaglowka: That's the one used context menu display. Could have left a comment..

PR #151, opened 2018-07-12
PP 0.2 report menu fix and refactor

glowka:


glowka:
No problem, I'll fix this modal right away

PR #163, opened 2018-09-03
GA events: annotation and extension-related

### LINKED ISSUE #160: Implement annotation-related GA events
Annotation-related events are in _Annotation_ section.
https://docs.google.com/spreadsheets/d/1QKMCeEtSoQ8UQt2VCmVjhWscd36nK5fv0RiJbOPzjmw/edit#gid=0

- [x] first message based events
- [x] configure GA custom dimensions
- [x] annotation display/click events
- [x] annotation add/edit events
- [x] annotation upvote events
- [x] annotation report events

### LINKED ISSUE #161: Implement popup-related GA events
Events:

- extensionDisabledOnAllSites
- extensionEnabledOnAllSites
- extensionDisabledOnSite
- extensionEnabledOnSite

from:

https://docs.google.com/spreadsheets/d/1QKMCeEtSoQ8UQt2VCmVjhWscd36nK5fv0RiJbOPzjmw/edit#gid=0

glowka:
Resolves #160, resolves #161

### DIFF
--- a/src/pp-ga/core.ts
+++ b/src/pp-ga/core.ts
@@ -1,0 +1,41 @@
+import './ga.js';
+import FieldsObject = UniversalAnalytics.FieldsObject;
+import packageConf from '../../package.json';
+
+const GA_ID_PROD = 'UA-123054125-1';
+const GA_ID_DEV = 'UA-123054125-2';
+
+export const GACustomFieldsIndex = {
+  eventUrl: 'dimension1',
+  triggeredBy: 'dimension2',
+  priority: 'dimension3',
+  reason: 'dimension4',
+  isCommentBlank: 'dimension5',
+  annotationId: 'dimension6',
+  annotationLink: 'dimension7',
+};
+
+export function init() {
+  ga('create', PP_SETTINGS.DEV ? GA_ID_DEV : GA_ID_PROD);
+  // Our extension protocol is chrome which is not what GA expects. It will fall back to http(s)
+  ga('set', 'checkProtocolTask', () => { /* nothing */ });
+  ga('set', 'appName', 'PP browser extension');
+  ga('set', 'appVersion', packageConf.version);
+}
+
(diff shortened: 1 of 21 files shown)


src/pp-ga/core.ts
```diff
@@ -0,0 +1,41 @@
+import './ga.js';
+import FieldsObject = UniversalAnalytics.FieldsObject;
+import packageConf from '../../package.json';
+
+const GA_ID_PROD = 'UA-123054125-1';
+const GA_ID_DEV = 'UA-123054125-2';
+
+export const GACustomFieldsIndex = {
+  eventUrl: 'dimension1',
```
kaglowka: Is it that the data should be presented to GA using these "dimension" labels?

glowka: GA has default dimensions (which are plenty) but are either basic and related to device, time etc or are more specialized but designed mainly for e-commerce. While we naturally, out-of-the-box use the first, we miss some more specific ones for our own purposes and here come _Custom Dimensions_ which are the way to extend default ones. In some other analytics system, in most cases paid, you do not have to declare like this, but that's how it goes with GA:
>Before you can send custom dimension and metric values to Analytics, they must first be defined in an Analytics property. Each Analytics property has 20 available indices for custom dimensions, and another 20 indices available for custom metrics.
>
>When you define a custom dimension or metric, you specify its name and other configuration values, and Analytics assigns an index number that you use later to reference the dimension or metric.

From https://support.google.com/analytics/answer/2709828?hl=en
Here is the list of the default dimensions: https://developers.google.com/analytics/devguides/reporting/core/dimsmets

glowka: One more thing is that GA requires events to have `category` and `action` dimensions, while `label` is optional, in fact `label` is really what matters and most systems have only this single dimension and __`label` is what we are used to call _event___.

PR #177, opened 2018-09-13
 Load annotation upvote on Viewer appearance + Some GA lost commits

### LINKED ISSUE #174: Load annotation upvote on Viewer appearance
Download upvote lazily for an annotation that is about to appear.

Major improvement: with current backend structure, upvote instance handling will be much cleaner.

glowka:
Resolves #174

### DIFF
--- a/src/components/viewer/Upvote.tsx
+++ b/src/components/viewer/Upvote.tsx
@@ -11,12 +11,15 @@ import {
 } from 'api/annotation-upvotes';
 import { PPScopeClass } from '../../class_consts';
 import ppGA from 'pp-ga';
+import { selectUpvote } from '../../store/api/selectors';
 
 interface IUpvoteProps {
   indirectChildClassName: string;
 
   annotation: AnnotationAPIModel;
-  upvote: AnnotationUpvoteAPIModel;
+  upvoteId?: string;
+  upvoteUrl: string;
+  upvote?: AnnotationUpvoteAPIModel;
 
   fetchUpvote: (url: string) => Promise<object>;
   deleteUpvote: (instance: AnnotationUpvoteAPIModel) => Promise<object>;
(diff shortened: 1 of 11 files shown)


src/components/viewer/Upvote.tsx
```diff
@@ -14,19 +14,27 @@ import ppGA from 'pp-ga';
...
   annotation: AnnotationAPIModel;
+  upvote: AnnotationUpvoteAPIModel;
 
+  fetchUpvote: (url: string) => Promise<object>;
   deleteUpvote: (instance: AnnotationUpvoteAPIModel) => Promise<object>;
   createUpvote: (instance: AnnotationUpvoteAPICreateModel) => Promise<object>;
 }
 
 interface IUpvoteState {
-  initialView: boolean; // used to determine whether edit/delete buttons should be visible
```
kaglowka: This field was used by edit/delete buttons to determine whether they are still visible, I suppose that's an unintended change :)

kaglowka: (they disappear after some time when the mouse is not over an annotation any more)

glowka: It was obsolete code after file splitting, this field is used elsewhere :P

src/components/viewer/Upvote.tsx
```diff
@@ -35,23 +43,22 @@ export default class Upvote extends React.Component<Partial<IUpvoteProps>, Parti
 
   constructor(props: IUpvoteProps) {
     super(props);
+
+    if (props.annotation.relationships.annotationUpvote.data && !props.upvote) {
+      props.fetchUpvote(props.annotation.relationships.annotationUpvote.links.related).then(
```
kaglowka: So slick :+1:  :)

Issue #175, opened 2018-09-12
Save more location information to API on annotation creation


kaglowka:
**make sure we provide**

- [x] `quote` -  the exact annotated text

**provide**

- [ ] `quoteContext` - the wider textual context for future use.

PR #185, opened 2018-09-23
Pass quote text to backend when creating annotation

glowka:
Always sending 'quote` when creating annotation resolves part of #175 and brings back the annotation adding function by adjusting to current backend requirements.

### DIFF
--- a/src/init/highlights.ts
+++ b/src/init/highlights.ts
@@ -41,10 +41,10 @@ function drawHighlights() {
     const annotationsToDraw = annotations.map((annotation) => {
       const { quote, range } = annotation.attributes;
       let locatedRange;
-      if (quote) {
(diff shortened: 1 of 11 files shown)


src/init/highlights.ts
```diff
@@ -41,10 +41,10 @@ function drawHighlights() {
     const annotationsToDraw = annotations.map((annotation) => {
       const { quote, range } = annotation.attributes;
       let locatedRange;
-      if (quote) {
-        locatedRange = uniqueTextToXPathRange(quote, document.body);
-      } else {
+      if (range) {
         locatedRange = range;
+      } else {
+        locatedRange = uniqueTextToXPathRange(quote, document.body);
```
kaglowka: Thanks for correcting!

PR #198, opened 2018-10-03
A fix to correctly detect empty annotation range (API returns an empty dictionary rather than a null/undefined value)

kaglowka:
The server currently returns an empty dictionary (literally, a JSON equal to `{}`) which is parsed to an empty  Object which evaluates to `true`. It would be more elegant if we returned a truly empty value from the server.
For now this will keep the extension working.

### DIFF
--- a/src/init/highlights.ts
+++ b/src/init/highlights.ts
@@ -41,7 +41,10 @@ function drawHighlights() {
     const annotationsToDraw = annotations.map((annotation) => {
       const { quote, range } = annotation.attributes;
       let locatedRange;
-      if (range) {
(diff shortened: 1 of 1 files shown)


kaglowka:
Sure, it was meant as a fix for several hours until backend is fixed (master was broken). If you can fix backend right away, go ahead! I made the quickest and laziest solution so it could be merged if needed.

glowka:
I meant: it's a one-liner, so just fix backend yourself :)

kaglowka:
Oh well ;)
https://github.com/PrzypisPowszechny/pp/pull/61 fixes this issue on the server side and requires no modifications on the client side.

Issue #199, opened 2018-10-03
A major src directory refactor


kaglowka:
- clear separation of different extension components (popup, content scripts, etc.)
- more structured application core (currently the most of the application logic is contained in `src/init` directory)

PR #201, opened 2018-10-05
Annotation data flow modifications + additional location information + docs

### LINKED ISSUE #175: Save more location information to API on annotation creation

### LINKED ISSUE #183: An intermediate step in annotation processing
An additional component subscribing to store changes; when the store changes, the component tries to locate the annotations within text. When it succeeds, it saves them to `locatedAnnotations` state field. Otherwise it saves them in `unlocatedAnnotations`.
All other components dependent on annotations consume data from `locatedAnnotations` and `unlocatedAnnotations` rather than directly from `state.api`.

kaglowka:
Resolves #175 
Resolves #183 

- TextSelector uses only pure `xpath-range` ranges data and is no longer involved in any application logic
- `AnnotationLocation` -- a new explicit grouping of location-related data saved and received with an annotation 
- intermediate explicit annotation location step between data loading and attaching to DOM -- all annotations are explicitly sorted into those that could be located and those that could not.
- Basic information on the DOM annotation-related libraries used in the application
- additional related refactors

This PR should be followed by directory refactor in the source root ( #199 )

### DIFF
--- a/src/components/menu/Menu.tsx
+++ b/src/components/menu/Menu.tsx
@@ -17,7 +17,7 @@ interface IMenuProps {
   locationX: number;
   locationY: number;
   range: XPathRange.SerializedRange;
-  text: string;
+  quote: string;
 
   setSelectionRange: (range: AnnotationLocation) => void;
   showEditor: (x: number, y: number) => void;
@@ -35,7 +35,7 @@ interface IMenuProps {
     locationX,
     locationY,
     range: state.textSelector.range,
-    text: state.textSelector.text,
+    quote: state.textSelector.quote,
   };
 },
   {
@@ -61,11 +61,11 @@ export default class Menu extends React.Component<Partial<IMenuProps>, {}> {
       locationX,
       locationY,
       range,
-      text,
+      quote,
     } = this.props;
 
     this.props.hideMenu();
-    this.props.setSelectionRange({range, text});
+    this.props.setSelectionRange({ range, quote });
     this.props.showEditor(locationX, locationY);
     ppGA.annotationAddFormDisplayed('addingModeMenu');
   }
--- a/src/components/menu/Menu.tsx
+++ b/src/components/menu/Menu.tsx
@@ -16,8 +16,7 @@ import { AnnotationLocation } from '../../utils/annotations';
 interface IMenuProps {
   locationX: number;
   locationY: number;
-  range: XPathRange.SerializedRange;
-  quote: string;
+  annotationLocation: AnnotationLocation;
 
   setSelectionRange: (range: AnnotationLocation) => void;
   showEditor: (x: number, y: number) => void;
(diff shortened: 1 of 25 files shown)


glowka:
At first glance looks like great changes, with extensive docs and stuff, but I'll go through it more thoroughly tomorrow with fresh mind :)

> 
<sub>[View in Slack](https://przypispowszechny.slack.com/archives/CCDNT60UF/p1538757724000100)</sub>

src/components/menu/Menu.tsx
```diff
@@ -8,18 +8,17 @@ import Widget from 'components/widget';
...
 
 interface IMenuProps {
   locationX: number;
   locationY: number;
-  range: Range.SerializedRange;
-  text: string;
+  annotationLocation: AnnotationLocation;
 
-  setSelectionRange: (range: SerializedRangeWithText) => void;
+  setSelectionRange: (range: AnnotationLocation) => void;
```
glowka: How about going for it and updating `range` var to something? What do you think?

src/core/Highlighter.ts
```diff
@@ -1,6 +1,6 @@
 import $ from 'jquery';
 import { PPHighlightClass } from 'class_consts';
-import { Range } from 'xpath-range';
+import { Range as XPathRange } from 'xpath-range';
```
glowka: So simple, yet so nice

src/init/processAnnotations.ts
```diff
@@ -0,0 +1,62 @@
...
+import { findUniqueTextInDOMAsRange } from '../utils/annotations';
+import { locateAnnotations } from '../store/annotations/actions';
+import { LocatedAnnotation } from '../store/annotations/types';
+import { AnnotationAPIModel } from '../api/annotations';
+
+let instance;
+
+function init() {
+  // subscribe to store changes and return unsubscribe fn
+  const unsubscribe = store.subscribe(processAnnotations);
```
glowka: Hmm, listening to every change of store(subscribe) and then for any change, locate all annotations. I know it was worse than it is now, but sooner or later I would aim for reducers instead of subscriptions (attached to three actions API_CREATED, API_READ, API_UPDATED) and components tree instead of "manual" drawing of highlights. Then react+redux can optimize re-rendering and the whole flow will be more under control.

src/components/menu/Menu.tsx
kaglowka: Oh yes, surely. I just had to stop somewhere :)

kaglowka: I'll improve on that very soon.

src/init/processAnnotations.ts
kaglowka: As we agreed at the meeting, this should be eventually done. I opened #209

Issue #70, opened 2018-05-25
Basic automatic browser extension build config


kaglowka:
- [x] put manifest.json in dist
- [x] build popup and include it in manifest (browser action default_popup)
- [x] put icon in dist

PR #220, opened 2018-10-13
Add crucial information in HTTP headers: site URL and CSRF token + context menu fix

### LINKED ISSUE #176: Modify PP GET annotation requests to correspond the backend privacy improvements
Solution architecture in backend issue: https://github.com/PrzypisPowszechny/pp/issues/39

### Frontend details

We need to  create `readPostEndopint` basing on `readEndpoint` which will do the same as the latter, but using `POST` method. This means copying the body of it `readPoint` from https://github.com/redux-json-api/redux-json-api/blob/9e4210af054e3ccdc61c312480de09ba844760fd/src/jsonapi.js#L85 and adding just few lines.

Beside that we could create the issue in https://github.com/redux-json-api/redux-json-api/ and ask for adding options to `readEndpoint` which would be passed to the request, but that library is very specialized and we cannot be certain it will be accepted.

### LINKED ISSUE #218: Add sending X-CSRFToken header with every request
Using this hook: https://github.com/axios/axios#interceptors

This issue should be completed before merging: https://github.com/PrzypisPowszechny/pp/pull/70

kaglowka:
Resolves #176 
Resolves #218 

CSRF token is available via chrome extension background page `chrome.cookies` API, since traditional `document.cookie` won't return the cookies set by HTTP requests since the API domain and background-page/content script domain differ.

The already complicated webpack configuration got temporarily slightly more complicated and simplifying it should be a priority.

After merging, merge https://github.com/PrzypisPowszechny/pp/pull/70

### DIFF
--- a/src/content-scripts/init-API.ts
+++ b/src/content-scripts/init-API.ts
@@ -2,11 +2,39 @@ import store from './store/store';
 import axios from 'axios';
 import { setAxiosConfig } from 'redux-json-api';
 
-export function configureAPIRequests() {
+function getExtensionCookie(name: string) {
+  // Read special per-extension cookie
+  // available not for the host domain (unlike traditional website cookies) but for this particular extension client
+  return new Promise((resolve) => {
+    chrome.runtime.sendMessage({ action: 'GET_COOKIE', name }, (response) => {
+      if (response) {
+        resolve(response.value);
+      } else {
+        resolve(null);
+      }
+    });
+  });
+}
 
-  // add header consumed by '/annotations' endpoint for URL filtering
+export function configureAPIRequests() {
+  /*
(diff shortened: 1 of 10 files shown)


src/content-scripts/init-API.ts
```diff
@@ -0,0 +1,46 @@
+import store from './store/store';
+import axios from 'axios';
+import { setAxiosConfig } from 'redux-json-api';
+
+function getExtensionCookie(name: string) {
```
glowka: Nice! I am gonna use it with ppGA soon

kaglowka: I didn't export it just yet, but I thought of it as a util-to-be ;)

PR #244, opened 2018-11-17
Selenium env fixes

kaglowka:
Two bug fixes

- `csrftoken` was expected for any request method (not only for modification requests) and `csrftoken` cookie was expected while making any request
- cookie request from content script to background script caused crash when running extension in chrome with selenium; as a consequence, content script axios fetches blocked waiting synchronously for response from the background script


glowka:
Quite vicious bugs :D Great to have them fixed!

One more thing: I will change the name of the PR since it is more about cookies/csrf than selenium.

PR #254, opened 2018-12-03
Add saving annotation requests from popup and context menu

### LINKED ISSUE #234: Add annotation request saving (mail / database) on button click

### LINKED ISSUE #238: Add "annotation request" in context menu

kaglowka:
Resolves #234 
Resolves #238


src/common/api/annotation-requests.ts
```diff
@@ -0,0 +1,14 @@
+
+// Annotation request is not fully a JSON API model (it misses type id)
+// (besides the missing "type" field the schema is compliant)
```
glowka: But wait: why haven't we actually made it fully JSON API compliant?

kaglowka: We could mock `id` and increment it + provide JSON-API resource type. I see no point in it, since we, you and I, are aware this is just a "send-an-e-mail" endpoint and doing it a little bit JSON-API like was the bit I had to do anyway somehow and I chose to do it JSON-API way. I've done the most to use our standard utils (deserializers etc.) in the backend to be able to switch easily once we actually persist this data elsewhere than just sending an email.
I'd rather not do such mocks for the feature we have judged experimental. I'd rather update both the client and the server once we do need it JSON-API compliance.

glowka: But we stick to JSON-API which says if you add a resource to the server you get object with unique type+id, so what's wrong with endpoint returning both type and id, with the later being for example uuid?
This way data posted and received is exactly as it is supposed to be: you post and object and get unique object. I do not see it as mocking in any way. I know that we are not going to do any operations with received object, but JSON-API does not obligate you to support all operations at the front-end site. So if our main method is 100% supporting that case without any design complications, why are we moving towards dualism of main method (redux-json-api way) and one introduced here?

We do not have to agree on this at this moment, I am happy enough about the views exchange, so I am approving the PR.

kaglowka: Theoretically I agree -- I just didn't choose to do that one more step since I found it unnecessary. But i don't think that adding 2 fields is worth such a length discussion, too ;)

PR #270, opened 2019-01-02
Improve `npm start` by automatically open Chrome and loading the extension

kaglowka:


glowka:
Can we add an option to choose between chromium and chrome when calling `npm start`?

Or  alternatively @kaglowka @KasiaSkowronska can we switch to using chromium as development browser?

It is identical as chrome (actually chrome is just chromium with changed branding) and using it for development has some additional advantages: 
- developer of PP extension that normally uses chrome as standard browser can use "real" extension installed from the store, no need to switch on/off production and development instances when all development is on chromium
-when developing on chromium only necessary browsers are reloaded, all personal tabs remain untouched as they are in separate browser
- chromium is ideal development environment: no other extension, no issues with browser running prior to calling `npm start` as nobody uses it ;)

What do you say?

kaglowka:
:100: 
Yes and yes :)
I'll think of a way to select chromium / chrome, but actually I'm all for chromium anyway :)

kaglowka:
Finally I simply replaced Chrome with Chromium for now.

PR #275, opened 2019-01-06
Annotation list in popup

### LINKED ISSUE #98: A website "blacklist" to be ignored by the extension.
Annotating most of the websites does not make sense. We could easily "blacklist" the most popular ones (facebook, slack, etc.)
- if they are JS-heavy, we avoid the risk of interfering with their scripts and/or causing performance problems.
- if the user for some reason finds it reasonable to annotate one of those sites, they can be annoyed by the annotations disappearance (sure to happen on very dynamic websites like facebook etc.)
- when the user is notified that a website XXX cannot be annotated, it can improve his understanding of PP's purpose.

Currently, several popular sites are blacklisted in `manifest.json`.
- [ ] explicitly inform the user that the current website is blacklisted
- [ ] reasonably extend the black list

kaglowka:
To be merged only after #274 

Resolves #98 (out of the box).
When the website content script is inactive, the popup now informs the user that the website is not supported.
It's brutal, but also very effective. We can freely extend website blacklist in the extension manifest and the popup will pick it up, that the website does not cooperate with it.

### Suggested follow-up:
- rename BrowserPopupNavigator / BrowserPopup so they express better the hierarchy


glowka:
oh, I have just realized no review was requested here only after I rebased it :upside_down_face:

kaglowka:
It's ok :)
I didn't do it since I wanted to discuss first what kind of feature this is.
;)

glowka:
Were the toggle buttons supposed to look like that after the changes?

screenshot from 2019-01-10 18-47-48

kaglowka:
No, definitely toggle buttons shouldn't look like that, I'll check it ;)

kaglowka:
I'm merging this despite the error so I can also rebase #276

PR #287, opened 2019-01-13
Annotation summary styling

### LINKED ISSUE #283: Add styling to popup AnnotationSummary
There is no styling of AnnotationSummary at all.

Additionally, with this new component the overall content of popup exceeds its size, in result there is scrollbar present on the right of popup.

KasiaSkowronska:
Resolves #283

### DIFF
--- a/src/popup/components/annotationList/AnnotationList.tsx
+++ b/src/popup/components/annotationList/AnnotationList.tsx
@@ -1,9 +1,11 @@
 import React from 'react';
 import { loadAnnotationLocationData, PopupAnnotationLocationData, sendScrollToAnnotation } from '../../messages';
 import { PopupPages } from '../BrowserPopupNavigator';
+import { AnnotationPPCategories } from 'common/api/annotations';
 import styles from './AnnotationList.scss';
 import { Icon } from 'react-icons-kit';
 import { ic_chevron_left } from 'react-icons-kit/md/ic_chevron_left';
+import classNames from 'classnames';
 
 export interface IAnnotationListProps {
   onPageChange: (Event) => void;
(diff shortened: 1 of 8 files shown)


src/popup/components/annotationList/AnnotationList.tsx
```diff
@@ -49,15 +54,22 @@ export default class AnnotationList extends React.Component<Partial<IAnnotationL
     } else {
       return (
         <div className={styles.self}>
-          <button onClick={this.handleGoBackClick}>Wróć</button>
-          <ul>
+          <div onClick={this.handleGoBackClick}>
+            <Icon className={styles.chevronButton} icon={ic_chevron_left} size={25}/>
+          </div>
+          <ul className={styles.annotationList}>
```
kaglowka: It would be nice to ensure constant width for the case when the list is being loaded (a few lines above) and also include some basic styling for loading.

src/popup/components/annotationSummary/AnnotationSummary.tsx
```diff
@@ -5,7 +5,10 @@ import {
   AnnotationPPCategories,
   annotationPPCategoriesLabels,
 } from 'common/api/annotations';
-import _countBy from 'lodash/countBy';
+import _ from 'lodash';
```
kaglowka: The idea was to not import whole lodash and increase the package size.
I just checked that, at least with Webpack 4, it does not matter. We can always simply import `_ from 'lodash'`

src/popup/components/annotationList/AnnotationList.tsx
```diff
@@ -49,15 +54,22 @@ export default class AnnotationList extends React.Component<Partial<IAnnotationL
...
           {this.state.annotationLocationData.located.map(annotation => (
             <li
+              className={styles.listItem}
               key={annotation.id}
               data-annotation-id={annotation.id}
-              onClick={this.onAnnotationClick}
+              onClick={() => sendScrollToAnnotation(annotation.id)}
```
kaglowka: We decided to stick to a policy of no arrow functions as handlers. We might revise in the future, but for now we stick to it.

PR #308, opened 2019-02-16
WIP: Add logging in with social accounts

glowka:
This is demonstration how this should (or could) be implemented. It works but this is just a nice start for future assignee.


kaglowka:
Merged with #316

PR #312, opened 2019-02-25
Improve error handling in communication with background + misc

### LINKED ISSUE #289: Remove the abstraction over sync and local versions of chrome.storage
`local` and `sync` are used for, respectively, `dev` and `prod` builds.
This differentation might not be adequate, since e.g. usage of these two storage types within a build might be well justified.
It makes sense in as much as the `sync` storage type has low quotas (since it additionally involves a bit of Google infrastructure).
A better answer to quota problem might be to introduce a policy of economic storage updates.

kaglowka:
Resolve #289 
For now we use only `chrome.storage.local`. We might switch later to `sync` if we need any kind of data to be shared between browser instances.


glowka:
I guess these changes were rather preliminary and were followed by much thorough works, so the #313 is actually the proper one which will be merged soon, while this one can just be closed, right @kaglowka ?

kaglowka:
No, they are completely independent :) probably with minor conflicts

glowka:
But isn't this one containing some incomplete changes, e.g. mocking that
was fixed later on?

kaglowka:
No, I removed the dev client mock + there is no differentiation between the type of storage used in dev and production (all use `chrome.storage.local` and never `chrome.storage.sync`).

The next PR adds chrome API mock in `jest` tests.

PR #339, opened 2019-03-30
Popup communication

kaglowka:
To be merged after #338 

Resolve #330 
Resolve #323

Issue #314, opened 2019-03-02
Upgrade react to 16.4


kaglowka:
React 16.4 "fixed" `getDerivedStateFromProps` [here](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html). This is probably the reason updating React to 16.4 causes our Viewer to stop appearing.
We should update react when the time comes so we don't get stuck here with a temporary anti-pattern working in 16.3.

kaglowka:
Done in #357

PR #340, opened 2019-04-16
Set up background driven authorization

### LINKED ISSUE #325: Fix login button bug
The login button needs to be clicked again after successful login before the popup interface updates.
The error is most probably due to popup closing before the authorization is finished (and so the result is not handled).
It is deceiving that popup doesn't close with popup inspector open -- hence the bug.

glowka:
Resolves #325


glowka:
Some e2e tests seems a bit unstable, they sometimes pass, sometimes not. There must be some kind of issue with them. I need to check it before merging.

src/content-scripts/annotations/AnnotationLocator.ts
```diff
@@ -3,15 +3,15 @@ import 'rangy/lib/rangy-classapplier';
 import 'rangy/lib/rangy-highlighter';
 import 'rangy/lib/rangy-textrange';
 import 'rangy/lib/rangy-serializer';
-import { LocatedAnnotation } from '../../common/store/tabs/tab/annotations/types';
+import { LocatedAnnotation } from 'common/store/tabs/tab/annotations/types';
 import { annotationRootNode } from '../settings';
 import { Range as XPathRange } from 'xpath-range';
-import { escapeRegExp } from 'tslint/lib/utils';
```
glowka: @kaglowka I found out that importing from typescript or tslint is very costly. It leads to slowness of build and some webpack warnings  when source-map-support does not know how to handle _special_ files from typescript which normally are used to compiling not being compiled.
It's not a big deal for us, because we actually do not use that typescript code that trans-piles TS into JS, it was just a replaceable util in this case, but still enough to triggers import of massive codebase ;)

kaglowka: Good job! The compilation time is actually a great deal, with how much TS compilation takes along with extension reloading etc.

kaglowka:
The upgrade to React > 16.4 stops Viewer from appearing (#314)
I'll add a simple (and important) e2e test to detect problems like that.

kaglowka:
Since changing the logic of Viewer appearing is rather time-consuming and I'm sure we'd like to rework it sooner or later anyway, I guess it's better to stick to React 16.3 (and possibly other packages like e.g. `jest`) for now.

kaglowka:
I added the test to #339 (the test is passed at that point)

.babelrc
```diff
@@ -1,14 +1,14 @@
...
       "modules": false
     }],
-    "react"
+    "@babel/preset-react"
   ],
   "env": {
     "test": {
       "presets": [
-        ["es2015"], "react"
+        ["@babel/preset-env"], "@babel/preset-react"
```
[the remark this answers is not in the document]
glowka: babel migrated in the meantime:
 - from multiple `preset-esXXXX` packages  into universal `preset-env`
 - from `babel-package-name` name style into `@bable/package-name`

glowka:
I'll revert that react upgrade. 

Thanks a lot for the regression test!

glowka:
> Some e2e tests seems a bit unstable, they sometimes pass, sometimes not. There must be some kind of issue with them. I need to check it before merging.

BTW Upgrading stabilized the tests :)

glowka:
I rebased it on master with this new test – passing after your @kaglowka downgrade

src/popup/components/LoginForm.tsx
```diff
@@ -61,80 +36,15 @@ export default class LoginForm extends React.Component<Partial<LoginFormProps>,
...
-    authURL += `&response_type=token`;
-    authURL += `&redirect_uri=${encodeURIComponent(redirectURL)}`;
-    authURL += `&scope=${encodeURIComponent(scopes.join(' '))}`;
-
-    return chrome.identity.launchWebAuthFlow({
-      interactive: true,
-      url: authURL,
-    }, data => this.ppAuthenticate(data, 'google'));
+    this.props.initiateAuthProc(AuthProviders.google);
   }
```
kaglowka: Super cool :) A perfect presentation component :)

src/common/store/background-epics.ts
```diff
@@ -16,15 +30,53 @@ interface FluxStandardAction {
...
+        }),
+        map(newAction => syncTabMark(action, newAction)),
+      ),
+    ),
+  )
+);
+
+export const propagateAuthenticationDataEpic: Epic<FluxStandardAction, any, IState> = action$ => (
+  action$.pipe(
+    filter(action => [USER_DATA_NEW, USER_ACCESS_TOKEN_REFRESHED, USER_DATA_CLEARED].includes(action.type)),
```
kaglowka: Actually that's quite a nice way to declare relationship between these actions and login info propagation to dashboard :)

src/common/store/storage/actions-background.ts
```diff
@@ -8,20 +7,17 @@ import dashboardMessaging from '../../../background/dashboard-messaging';
...
     dispatch(userTokensRefreshed(auth));
-    dashboardMessaging.sendLoginData();
   };
 }
 
 export function userLoggedOut() {
   return (dispatch, state) => {
     dispatch(userDataCleared());
-    dashboardMessaging.sendLoginData();
   };
```
kaglowka: These thunks are no more needed and don't even need to be thunks. I'll remove them in one of the upcoming PRs

src/content-scripts/annotations/AnnotationLocator.ts
kaglowka: I just realized I imported it from tslint by accident :)

src/common/store/background-epics.ts
```diff
@@ -16,15 +30,53 @@ interface FluxStandardAction {
...
+    mergeMap( (action): Observable<FluxStandardAction> =>
+      from(authenticate(action.payload.provider)).pipe(
+        mergeMap( loginData =>
+          loginData ? of(
+            completeAuthProc(),
+            userDataNew(loginData),
+          ) : of(
+            cancelAuthProc('Anulowano autoryzację'),
+          ),
+        ),
```
kaglowka: One day I'll get round to understanding this :)

glowka: I tried to describe it, but gave up after seeing how much of RxJS docs I actually copy ;) We could talk about it for 10mins – it would be the fastest way :)

glowka:
@kaglowka Happy to read your immensely positive review :)

PR #341, opened 2019-04-17
Popup emulation for e2e testing & build acceleration

kaglowka:
Merge after #340 

- [x] exact package versionss
- [x] new test commands
- [x] `npm start` oopening page with shortcuts
- [x] enable popup testing parallel to the content script testing (within a single test case)
- [x] refactor popup emulation
- [x] fix e2e-build-optimized
- [x] add a basic popup modes test
- [x] fix test-unit
### optional
- [x] add some basic popup test using popup emulation
- [ ] add an npm command that returns non-zero code on any webpack warnings

Since this PR has grown rather huge, I think it can be merged right away, without fixing all the minor issues it generates. The remaining issues can be fixed later on without blocking development.

**follow up**
- [ ] fix new `npm start` extension switching off. Probably this is a new issue -- due to some errors I couldn't identify, the extension turns off in chromium. It must be then turned on manually in chromium settings. This is rather annoying, so should be solved quickly.
- [ ] refactor tests (constants etc.)
- [ ] refactor flow (variable names)
- [ ] add docs on popup emulation


README.md
```diff
@@ -177,6 +177,13 @@ _Out of operation for now._
 npm run build-client
 ```
 
+## Modifying packages
+We usually want to install the exact version of a package to ensure no differences between our development environments.
+You can set it the default behaviour when installing new package from commandline.
+```
+npm config set save-exact=true
```
glowka: Useful hint, I didn't know that config

config/dev.plugins.js
```diff
@@ -9,7 +9,7 @@ module.exports = [
   new ForkTsCheckerWebpackPlugin({
     async: true,
     watch: 'src',
-    tslint: true,
+    // tslint: true,
```
glowka: Is this intended?

dev-pages/util-page.html
```diff
@@ -0,0 +1,34 @@
...
+        and other relationships that a normal backgorund page/popup would have.
+        E.g. they exchange messages etc.
+        The only issue is their coexistence, i.e. unusual interaction of actions this may generate
+
+    </li>
+    <li> as long as the "forked" page does not interfere with the application logic, it can be well used for testing</li>
+    <li> e.g. in popup, when the tab does not modify the global application state spontaneously (e.g. via timer) everything should be fine</li>
+</ul>
+</body>
+</html>
```
glowka: :+1:

src/common/store/background-epics.ts
```diff
@@ -20,19 +20,21 @@ import {
...
-interface FluxStandardAction {
+export interface FluxStandardAction {
   type: string | symbol | any;
   payload?: any;
   error?: boolean | any;
   meta?: any;
 }
 
-export const locateEpic: Epic<FluxStandardAction> = action$ => action$.pipe(
+export type PPEpic = Epic<FluxStandardAction, FluxStandardAction, IState>
```
glowka: If it is general util,so let's call it in general way and reserve `PP` prefix for generating some unique namespace etc. What do you think about defining such cases of usage?

src/common/store/runtime/types.ts
```diff
@@ -18,11 +18,11 @@ export enum AuthProcStages {
...
 export interface IAuthProcState {
   stage: AuthProcStages;
-  provider: AuthProviders;
+  provider?: AuthProviders;
   msg?: string;
 }
 
 export interface IRuntimeState {
-  user: IUserState;
+  // user: IUserState;
```
glowka: Intended?

src/content-scripts/components/App.test.tsx
```diff
@@ -66,48 +71,49 @@ describe('App component', () => {
...
-          },
-          appModes: {
-            isExtensionDisabled: false,
-            disabledPages: [],
-          },
-        },
-      },
-    };
+    const initialState = { ...correctStoreState };
+    // const initialState = {
```
glowka: Uncomment/delete?

src/popup/popup.tsx
```diff
@@ -45,20 +43,78 @@ Promise.all([
     );
   });
 
-waitUntilStoreReady(store)
+waitUntilFirstStoreUpdate(store)
```
glowka: Nice concept of extending

 _module (DOM) loading  -> component first render_ 

to 

 _module (DOM) loading -> store ready -> store init action -> component first render._ 

Clear

README.md
kaglowka: :smile:

config/dev.plugins.js
kaglowka: No, thanks :)

src/common/store/background-epics.ts
kaglowka: Since this type is specific for our project (using `IState` I thought it was not universal enough). I'm thinking now of `StandardEpic`. What do you think?

src/common/store/runtime/types.ts
kaglowka: Actually I meant to remove it altogether, as we do not store any runtime user data, so I'd better just remove it for now.

src/content-scripts/components/App.test.tsx
kaglowka: Sure, thanks :)

src/common/store/background-epics.ts
glowka: Yep, much better :)

PR #357, opened 2019-05-23
Amend components and upgrade React to latest version

kaglowka:


src/content-scripts/components/AnnotationRequestForm/AnnotationRequestForm.tsx
```diff
@@ -54,21 +54,17 @@ interface AnnotationRequestFormState extends IAnnotationRequestFormData {
...
 
-  static getDerivedStateFromProps(nextProps, prevState) {
-    return {
+  constructor(props: AnnotationRequestFormProps) {
+    super(props);
+    this.state = {
       quote: '',
       comment: '',
-      ...nextProps.initialData,
+      ...props.initialData,
```
glowka: Is constructor called every time props changes?
I am asking becuse such was current implementation and changing it will break how form behaves if it has been already opened when request annotation action is called for another quote.

kaglowka: You're right, I forgot it can be updated (and not just mounted / unmounted)

glowka: If we want to get rid of derived props I would follow this `key={id}` (or `key={quote}` in this case`) solution that we once considered and it also in the [article](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#recommendation-fully-uncontrolled-component-with-a-key) you mentioned.

kaglowka: (and no, it is not)

kaglowka: I just thought that too so I'm going to do just that :)

kaglowka: It is fixed now

PR #351, opened 2019-05-09
Annotation request viewer

kaglowka:
To merge after #349 

Resolves #329 

- [x] loading and locating annotaiton requests
- [x] sync creating / deleting annotation requests with location
- [x] highlight annotation requests in text
- [ ] display annotation request viewer
- [ ] handle mouse events
- [ ] add annotation request viewer style
- [ ] add quote context

**Temporarily just detect click on an annotation request highlight** instead of displaying a viewer on highlight hover

PR #358, opened 2019-05-23
Annotation mouse events

kaglowka:
To be merged after #357 

Preparing for #351 (I just reopened it since it was not fully finished...)

Clearer division of responsibilities:
- highlight-manager only sets the viewer window open or reopens it when hovering over other highlights
- Viewer sets redux mouseOver state
- ViewerManager tracks changes in redux state and decides whether to close the window

This is meant as a solid base from which annotationRequest highlighting (almost trivial compared to Viewer) can be replicated.


glowka:
I rebased it on current master

src/content-scripts/components/viewer/ViewerManager.tsx
```diff
@@ -6,40 +6,38 @@ import _isEqual from 'lodash/isEqual';
...
+  isDeleteModalOpen: boolean;
+  annotationIds: string[];
+  isAnyReportEditorOpen: boolean;
+
+  hideViewer: () => void;
+  setMouseOverViewer: (value: boolean) => void;
+}
+
 interface IViewerManagerState {
-  isMouseOver: boolean;
```
glowka: We have now plenty of other variables starting with `is...`, so are you sure you want to make an exception?

I thought about this is/has convention for a second if it is only me liking it or it is maybe broader standard (still I don't mean _universal_). So I googled and it looks like the latter, for example there is event [eslint package](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/boolean-prop-naming.md) for that with 5k stars :smiley: so

kaglowka: I just wanted to be consistent (with `mouseOver` elsewhere), but we might make it consistent also the other way around :)

glowka: Oh, OK. I would prefer it this way :)

 We could also see how far we can propagate this convention.

kaglowka: Yeah, it might be fifty-fifty now for all I know :)

kaglowka:
> Hard for me to judge the details, but I will always support spiting responsibilities ;)

:) One more detail I forgot -- now viewer mouseOver state is only based on timer hover checking (rather than mouseenter / mousleave events)
This also removes redundancy.

At first I tried a converse approach, i.e. using only DOM mouseenter / mouseleave events, which was more or less successful **except for viewer non-DOM children that were mounted after Viewer first render**. It was tempting, since it didn't require timers, but all in all timer-based hover checks are probably more robust.

PR #360, opened 2019-05-24
Major popup fixes

kaglowka:
Merge after #358 (changes are basically independent, but there are some import conflicts)

- fix popup message "no annotation" appearing to early
- more detailed loading stage information in store
- corresponding loading stages of annotation requests
- different badge appearing during annotation location
- fix details of popup menu buttons clickability

As noted in the comments, two epics are now a bit redundant but I didn't want to focus on how to achieve the same in a more elegant way (with a lot of work still ahead of us...)

### DIFF
--- a/src/common/store/background-epics.ts
+++ b/src/common/store/background-epics.ts
@@ -61,7 +61,7 @@ export const annotationLocateEpic: StandardEpic = (action$, state$) => action$.p
 // TODO better return more than one action in annotationLocateEpic (not sure how to do it)
 export const annotationStageEpic: StandardEpic = (action$, state$) => action$.pipe(
   ofType(LOCATE_ANNOTATIONS),
-  map(action => {
+  map((action) => {
       const newAction = setAnnotationStage(AnnotationsStage.located);
       return syncTabMark(action, newAction);
     },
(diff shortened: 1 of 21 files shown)


src/common/store/background-epics.ts
```diff
@@ -50,11 +58,21 @@ export const annotationLocateEpic: StandardEpic = (action$, state$) => action$.p
...
+// TODO better return more than one action in annotationLocateEpic (not sure how to do it)
+export const annotationStageEpic: StandardEpic = (action$, state$) => action$.pipe(
+  ofType(LOCATE_ANNOTATIONS),
+  map((action) => {
+      const newAction = setAnnotationStage(AnnotationsStage.located);
+      return syncTabMark(action, newAction);
+    },
+  ),
+);
+
```
glowka: This task to emit more than one action looked interesting, so I solved it now:
```
export const annotationLocateEpic: StandardEpic = (action$, state$) => action$.pipe(
  ofType('API_READ'),
  filter(action => getActionResourceType(action) === resourceTypes.ANNOTATIONS),
  mergeMap(action =>
    defer(async () => {
        const tabId = retrieveLogicalActionTab(action, state$.value.tabs);
        setBadgeLocating(tabId);
        const locationData = await tabLocateAnnotations(tabId, action.payload.data);
        syncBadgeWithAnnotations(locationData, tabId);
        return [
          locateAnnotations(locationData), 
          setAnnotationStage(AnnotationsStage.located)
        ].map(
          newAction => syncTabMark(action, newAction)
        )
      },
    )
  ),
  // mergeMap treats array as a stream, every element is emitted separately
  mergeMap(action => action)
);
```

kaglowka: Thanks, that's cool!

kaglowka:
@glowka I used your solution + committed one more fix

PR #363, opened 2019-05-27
Add missing rrequest relation in AnnotationForm

glowka:


glowka:
@kaglowka As for now I haven't removed old editor yet. But I would go for it another PR, how do you think?

